PR #4, opened 2021-05-18
Update readme with correct spec link

mprorock:
Specification link was still pointing to previous link


OR13:
@mprorock sorry mike I pushed straight to main, didn't notice these changes. Do you think this PR is still necessary?

Issue #11, opened 2021-05-20
PROPOSAL: Define Client Credentials for APIs


OR13:
HTTP API integrations between trusted parties are a solved problem and do not require a new spec to provide support for the authorization we need.

- https://auth0.com/machine-to-machine
- https://developer.okta.com/docs/guides/implement-client-creds/overview/

These "Client Credentials" solutions are appropriate starting point for cross trust domain VC HTTP API operations.

This ticket seeks to gain consensus on their use with the vc http api.

For the sake of simplicity, lets assume 2 vendors wish to exchange presentations over https.

Vendor A -> Client Credentials App for Vendor B -> Vendor B uses app to access Vendor A's API.
Vendor B -> Client Credential App for Vendor A -> Vendor A uses app to access Vendor B's API.

OAuth client credentials setup is out of scope, but let us assume that each integration is identified with a CLIENT_ID and authenticated via a CLIENT_SECRET.

Before authenticated exchanges can be made, an application `access_token`'s for each app must be obtained using the client credentials.

Vendor A's software integration uses the client credentials from Vendor B's client credentials app and vendor b's IDP to obtain an access token, see these docs for the standard way this is accomplished: [auth0](https://auth0.com/blog/using-m2m-authorization/)

```
curl --location --request POST 'https://vendor-b.auth0.com/oauth/token' \
--header 'Content-Type: application/json' \
--data-raw '{
"client_id": "A's integration app for B's API",
"client_secret": "[You application client secret]",
"audience": "https://vendor-b.example.com",
"grant_type": "client_credentials"
}
```

once Vendor A' software has an `access_token` for Vendor B's API, the presentation exchange can begin, the flow is as follows:

###  Getting a Challenge for a flow

```
curl POST -H "Authorization: <ACCESS_TOKEN>" \
http://www.vendor-b.example.com/organizations/123/presentations/available  \
-d @./get-challengge-token-with-query-by-frame.json
```

```json
{
  "query": [
    {
      "type": "RequestQueryByFrame",
      "credentialQuery": [
        {
          "type": [
            "VerifiableCredential",
            "CommercialInvoiceCertificate"
          ],
          "reason": "Wallet XYZ is ready to selectively disclose new credentials."
        }
      ]
    }
  ]
}
```

Response is expected to look like this:

```json
{
  "query": [
    {
      "type": "QueryByFrame",
      "credentialQuery": {
        "frame": {
          "@context": [
[...]

PR #15, opened 2021-06-08
Refine Authorization Section

OR13:
- [x] Link to RFC RFC6749
- [x] Link to Okta
- [x] Link to Auth0

### DIFF
--- a/docs/sections/authorization.html
+++ b/docs/sections/authorization.html
@@ -30,7 +30,7 @@
 
     <li>
       <a href="https://auth0.com/machine-to-machine"
-        >Simplify Communication Between APIs and Trusted Services - Auth0</a
+        >Auth0 &mdash; Simplify Communication Between APIs and Trusted Services</a
       >
     </li>
   </ul>
--- a/docs/sections/authorization.html
+++ b/docs/sections/authorization.html
@@ -1,7 +1,32 @@
 <section id="authorization" class="normative">
   <h2>Open API Authorization</h2>
   <p>
-    Implementations MUST support Open ID Connect and MAY support HTTP Signatures
-    with Authorization Capabilities.
+    Implementations MUST support OAuth 2.0 Client Credentials based
+    authorization.
   </p>
+
+  <p>
+    See these links for more information on Machine to Machine API
+    authorization:
+  </p>
+
+  <ul>
+    <li>
+      <a href="https://datatracker.ietf.org/doc/html/rfc6749#section-4.4">
+        Client Credentials Grant - RFC6749
+      </a>
+    </li>
+    <li>
+      <a
+        href="https://developer.okta.com/docs/guides/implement-client-creds/overview/"
+        >Client Credentials flow - Okta</a
+      >
+    </li>
+
+    <li>
+      <a href="https://auth0.com/machine-to-machine"
+        >Simplify Communication Between APIs and Trusted Services - Auth0</a
+      >
+    </li>
+  </ul>
 </section>
--- a/docs/sections/authorization.html
+++ b/docs/sections/authorization.html
@@ -18,7 +18,7 @@
   <ul>
     <li>
       <a href="https://datatracker.ietf.org/doc/html/rfc6749#section-4.4">
-        Client Credentials Grant - RFC6749
+        RFC6749 &mdash; Client Credentials Grant
       </a>
     </li>
     <li>
--- a/docs/sections/authorization.html
+++ b/docs/sections/authorization.html
@@ -6,6 +6,11 @@
   </p>
 
   <p>
+    Client Credentials Grant is appropriate for verifying trusted application
+    authorization for Internal/External APIs.
+  </p>
+
+  <p>
     See these links for more information on Machine to Machine API
     authorization:
   </p>
--- a/docs/sections/authorization.html
+++ b/docs/sections/authorization.html
@@ -24,7 +24,7 @@
     <li>
       <a
         href="https://developer.okta.com/docs/guides/implement-client-creds/overview/"
-        >Client Credentials flow - Okta</a
+        >Okta &mdash; Client Credentials flow</a
       >
     </li>
 


docs/sections/authorization.html
```diff
@@ -1,7 +1,37 @@
 <section id="authorization" class="normative">
   <h2>Open API Authorization</h2>
   <p>
-    Implementations MUST support Open ID Connect and MAY support HTTP Signatures
-    with Authorization Capabilities.
+    Implementations MUST support OAuth 2.0 Client Credentials based
+    authorization.
```
TallTed, line 5: Are "Open ID Connect" and/or "HTTP Signatures with Authorization Capabilities" officially no good now?  Or should they be MAY additions to "OAuth 2.0 Client Credentials based authorization"?

docs/sections/authorization.html
```diff
@@ -1,7 +1,37 @@
...
+
+  <p>
+    See these links for more information on Machine to Machine API
+    authorization:
+  </p>
+
+  <ul>
+    <li>
+      <a href="https://datatracker.ietf.org/doc/html/rfc6749#section-4.4">
+        Client Credentials Grant - RFC6749
```
TallTed, line 21:
```suggestion
        RFC6749 &mdash; Client Credentials Grant
```

docs/sections/authorization.html
```diff
@@ -1,7 +1,37 @@
...
+      </a>
+    </li>
+    <li>
+      <a
+        href="https://developer.okta.com/docs/guides/implement-client-creds/overview/"
+        >Client Credentials flow - Okta</a
```
TallTed, line 27:
```suggestion
        >Okta &mdash; Client Credentials flow</a
```

docs/sections/authorization.html
```diff
@@ -1,7 +1,37 @@
...
+      >
+    </li>
+
+    <li>
+      <a href="https://auth0.com/machine-to-machine"
+        >Simplify Communication Between APIs and Trusted Services - Auth0</a
```
TallTed, line 33:
```suggestion
        >Auth0 &mdash; Simplify Communication Between APIs and Trusted Services</a
```

TallTed (changes requested):

docs/sections/authorization.html
OR13: OIDC does not make sense here, since we are trying to describe "Authorization" not "Authentication", there have been confused conversations about this in a number of meetings.

I think we should start with a single required MUST, and modify only based on strong evidence that additional MAYs are necessary...

This approach is fairly common, for example:

https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms

See "Recommended+"

> "HTTP Signatures with Authorization Capabilities"

This one is problematic for 2 reasons.

1. [HTTP Signatures spec is still WIP (IETF)](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12)
2. [Authorization Capabilities is still WIP (W3C-CCG)](https://w3c-ccg.github.io/zcap-ld/)

I suspect the VC-HTTP-API may not be able to make strong assertions regarding authorization support, if folks want to do VC-HTTP-API + ZCAP-LD + HTTP Signatures, they are and will remain free to do so.

For this work item (traceability-interop), we are trying to make recommendations that are production ready / stable today, and not work in progress drafts / or based on limited implementations that are unstable or incompletely implement a draft.

This was discussed on the vc-http-api call, and is still being discussed on the public credentials mailing list.

Whatever the vc-http-api decides, we will have to be compatible with, but if they say they support 5 different authorization strategies, we cannot forward that recommendation and succeed.

Picking even 2 mutually non interoperable authorization frameworks would be bad for interop.

I'd like to start with one, especially based on the review of:

https://github.com/w3c-ccg/traceability-interop/issues/11

mprorock (approved):

docs/sections/authorization.html
TallTed: You need to know with whom you're dealing (authentication) before you check whether they have permission to do something (authorization) -- even if that authentication is sometimes done through a authorization flow, as described at [https://connect2id.com/learn/openid-connect](https://connect2id.com/learn/openid-connect) -- 

> The OpenID authentication request is essentially an OAuth 2.0 authorisation request to access the user's identity, indicated by an `openid` value in the `scope` parameter.

Also note, [NIH Login OpenID Connect (OIDC) Service Details](https://webcache.googleusercontent.com/search?q=cache:6Qk_xtecm2wJ:https://auth.nih.gov/docs/OIDC/+&cd=17&hl=en&ct=clnk&gl=us) (from Google's cache, because NIH has restructured their live site a bit) --

> The OpenID Connect protocol extends the OAuth 2.0 protocol to add an authentication and identity layer for application developers. Where OAuth 2.0 provides the application developer with security tokens to be able to call back-end resources on behalf of an customer; OpenID Connect provides the application with information about the end-user, the context of their authentication, and access to their profile information.

I am not finding anything about authentication anywhere else in `traceability-interop`, so I think it must be handled here, as an integral part of authorization.

OR13: @TallTed yes,

Mechanically, https://datatracker.ietf.org/doc/html/rfc6749#section-4.4 

Defines a way to authenticate a client.

The entity authenticating is a software application. 

Registration of that application is out of scope, but is defined by OAuth, and described in the related links for Okta and Auth0.

Because that registration is with the IDP, both the VC-HTTP-API and this API do not describe it, but you can see how it works here:

https://auth0.com/docs/tokens/management-api-access-tokens/create-and-authorize-a-machine-to-machine-application

Once the application exists and is registered with the AS, an access token can be requested.

Perhaps your question is what scopes should be required on the access token?

Thats a different problem, and one which would require us to first agree to the flow this PR describes, since without this PR you cannot assert anything regarding support for OAuth 2.0 in this work item.

See also https://auth0.com/blog/on-the-nature-of-oauth2-scopes/

> TL;DR: Scopes only come into play in delegation scenarios, and always limit what an app can do on behalf of a user: a scope cannot allow an application to do more than what the user can do.

As a starting point I think this is fine, here is the code we use to get an application access token today:

```
const getAccessToken = async (clientId, clientSecret) => {
  const payload = {
    client_id: clientId,
    client_secret: clientSecret,
    audience: "https://vc-http-api.example.com",
    grant_type: "client_credentials",
  };

  console.log(JSON.stringify(payload));
  let response = await axios({
    method: "post",
    url: "https://vendor-name.auth0.com/oauth/token",
    headers: { "content-type": "application/json" },
    data: payload,
  });
  const { access_token } = response.data;
  return access_token;
};
```

... note the absence of scopes.

I can't speak for all traceability api implementers, but our APIs are segmented such that users and applications can do the same things, when assigned the same permissions... we don't use scopes to manage that because of the complexity noted in the section of the linked documentation:

> The Authorization Server Isn’t (Usually) Omniscient

We don't want to tell our AS every time a user changes what a client_id can do on our system.

This PR is just the start of addressing OAuth 2.0 and the tighter we bind to access_token shape, the more brittle our AuthZ approach becomes, and the harder it will be to configured an AS to produce a token, however, we could get very fine grained in the future, see:

https://github.com/swagger-api/swagger-petstore/blob/master/src/main/resources/openapi.yaml#L222

I think its best not to bite into this all at once.

OR13: ping @mprorock @tplooker any comments on this approach? would you be interested in formulating an opinion on how scopes might be used in the vc-http-api, or do you intend to authorize applications by id internally as we do currently?

TallTed: @OR13 -- Toward better understanding of this in context, I wanted to look at the rendered document, but it doesn't seem to be set up for previews.  Any chance those bits can be flipped?

OR13: IIRC thats a feature of W3C work items, not CCG work items, and is powered by ECHIDNA, which is a magical and brittle beast... 

I have also wished for the same thing, its a feature of some static web hosting solutions like netlify... its possible there is a low cost github action approach that might be possible, but I don't know for sure.

TallTed: @OR13 -- I know ECHIDNA has been and is being used on other CG reports (perhaps not CCG at present), and is producing, for example, [this preview](https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/173.html) of [this PR](https://github.com/w3c/rdf-star/pull/173) for the [RDF-star task force](https://w3c.github.io/rdf-star/) within [the RDF-DEV CG](https://www.w3.org/community/rdf-dev/).

Perhaps @vsnt @wyc @kimdhamilton can help with getting those bits flipped, though I don't think they will work on existing PRs -- I believe the previews only get generated for PRs initiated after preview generation is turned on.

OR13: @TallTed I opened https://github.com/w3c-ccg/traceability-interop/issues/19 to track the request for PR Preview rendering with ECHIDNA.

TallTed: I guess that since the spec doesn't explicitly exclude support of Open ID Connect, HTTP Signatures with Authorization Capabilities, or other authentication or authorization systems, such support is permitted, given that OAuth 2.0 Client Credentials are also supported.

That'll do for my initial concern on this PR, and ECHIDNA previews will help with others going forward.

OR13: @TallTed thats right, and we may add other authorization recommendations later, specifically as may be required for https://github.com/w3c-ccg/traceability-interop/issues/17 if that proposal gets a lot of support...

I think we need to be careful not to "just do whatever the vc-http-api does"... when it comes to allowing lots of optionality, especially at the AuthZ layer, as that will harm interop.... at the same time, we need to be careful to put forward recommendations that are feasible today and well documented / supported... or we will be agreeing to the spec equivalent of "cold fusion reactor interoperability"... when "cold fusion" is impossible.

thats not to say that we shouldn't also work on "cold fusion" in the appropriate wg / work item.

This work item was created because we needed a way to secure http api's that were exchanging VC / VP with each other. 

If we get bogged down in "building an authorization system so we can... exchange VC / VP with each other using a brand new standard we just invented"... thats not in scope for this work item, and still being debated as "in scope" for the vc-http-api.

I suspect one solution for the debate going on in the vc-http-api would be to create a new WG / work item, specifically targeted at "replacing OAuth2.0"... I am not sure we should take such work as a "blocking dependency" for interop.... we can always support it later, if it actually gets published as a TR / standard.

mkhraisha (approved):

OR13:
We agreed to merge this on the call today.

PR #24, opened 2021-08-25
docs: abstract and examples

nissimsan:
As agreed yesterday, starting point for 
https://github.com/w3c-ccg/traceability-interop/issues/22
and also some examples, ref
https://github.com/w3c-ccg/traceability-interop/issues/2

@mkhraisha, thoughts on the direction I'm going?

PR #26, opened 2021-09-28
fix typos introduced in #24

TallTed:
I'm sure of `Presentation` vs `Presenting`.  

I'm not sure of `proving` vs `proofing` in this context.


mprorock (approved):

mprorock:
good catch!

OR13 (approved):

PR #33, opened 2021-10-22
Fix respec swagger rendering issue with try it out

OR13:
The try it out feature of swagger can help us easily test multi implementation interop directly from the spec.

In the future, a user can select a server, authenticate, and verify a credential, then do the same thing with another server... where each server is hosted at a different location.

However, it requires us to alter the way that schemas are embedded in respec.


mkhraisha (approved):

nissimsan:
Action @OR13

OR13:
Ball is in my court, I need to update the PR so it can be merged without conflicts

Issue #34, opened 2021-10-28
Need ability to get credentials for automated testing


mprorock:
We will need be able to store some OAuth2 credentials for service providers to be able to retrieve a bearer token to run integration tests using the trace vocab against various VC-API providers complying with this interop profile.  

These tests will run using [newman](https://github.com/postmanlabs/newman) from gh actions

@brownoxford can you take a task of documenting the right approach so that service providers could add credentials to be used with testing, might be as simple as gh secrets, but then we will need to figure out a storage process for a list of secrets that can be matched to a list of service providers to be tested, and document a process for securely sending that info over to an editor on this project so that the keys can be added

brownoxford:
If I understand everything correctly, we are talking about requiring providers to provide an endpoint implementing [RFC 6749 client credentials grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) and to honor access tokens issued by said endpoint. Automated testing will require both the endpoint (not sensitive) and the client credentials (sensitive) for each provider.

I am inclined to agree that [GitHub Encrypted Secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) is probably the way to go. I would like to propose the use of gpg to encrypt secrets in transit when submitting new values to editors, but am open to other suggestions.

Should I be placing the documentation for this in the top-level README.md of this repo?

mprorock:
+1 top level `README.md` would be the right place

nissimsan:
@brownoxford working on proof of context this week.

OR13:
We can close this when Ci is operating in a secure setup and tests are passing.

brownoxford:
This initial PR sets the framework for providing the necessary configuration to fetch a client credentials grant token, next steps include:
* Create CI workflow that sets `client_secret` environment variables before running postman tests
* Refactor postman tests to begin using bearer token
* Generate additional postman tests to ensure that endpoints fail if bearer token is missing/invalid

nissimsan:
Keeping open until at least one other party beyond mesur have implemented this. 
Transmute will make an attempt on this soon.

OR13:
We support this now.

PR #40, opened 2021-12-07
feat: add code to fetch client_credentials grant token

brownoxford:
Ref: w3c-ccg/traceability-interop#34


mprorock (approved):

tests/postman/package.json
```diff
@@ -22,6 +22,7 @@
     "commander": "^7.2.0",
     "is-url-superb": "^6.1.0",
     "ky-universal": "^0.9.1",
+    "md5": "^2.3.0",
```
OR13, line 25: sadge

mprorock: sha would be better

mprorock (changes requested):
Per discussion on call - will add auth into one newman call

brownoxford:
Will also add note to documentation that retrieval of auth token may trigger billable event.

tests/postman/package.json
brownoxford: Updated to replace MD5 with SHA256 from crypto-js module

tests/postman/README.md
```diff
@@ -130,6 +130,8 @@ To import the collections into Postman, follow the guide on [importing Postman d
 
 ## OAuth 2.0 Configuration
 
+Postman tests make use of OAuth 2.0 client credentials flow for authentication, which may result in a billable event for configured authentication providers.
```
TallTed, line 133:
```suggestion
_**NOTE:** Postman tests make use of OAuth 2.0 client credentials flow for authentication, which **may result in a billable event** for configured authentication providers._
```

OR13 (approved):

mprorock (approved):

mkhraisha (approved):

PR #45, opened 2022-01-17
Propose did:web discoverability

OR13:


mkhraisha:
To make sure I'm reading this right, the proposal is to:

1. do away with did-configuration
2. require that the did document MUST contain the following properties:  [also known as, assertion method, verification method and service](https://www.w3.org/TR/did-core/#did-document-properties)
3. specifies the endpoints that require auth

Is that accurate?

What is the rationale behind removing did-config, you mentioned it has some issues around the multi-tenant setup, would you mind expanding?

OR13:
> What is the rationale behind removing did-config, you mentioned it has some issues around the multi-tenant setup, would you mind expanding?

Assuming did configuration requires all API implementers adhere to this spec: 

https://identity.foundation/.well-known/resources/did-configuration/

And even then, it won't work for multi-tenant setups.

Discovering these features is the first step for automated interoperability testing.

All a vendor should need is a `did:web` and a `CLIENT_ID` and `CLIENT_SECRET` and the rest can be automated (without needing another draft spec).

We require a solution that supports multi-tenant setups. 

We are open to a counter proposal.

mkhraisha:
personally I'm happy to not require did config spec, less spec for interop makes it easier for everyone, 
I'm just not sure what the issues with multi-tenant support it has, can you link me to any issues to read?

OR13:
> I'm just not sure what the issues with multi-tenant support it has, can you link me to any issues to read?

You should read the did config spec, it only supports naked origins: `example.com` NOT `platform.example.com/org/456`

mkhraisha:
Oh I see. 
When we set it up for that spec our tenants were `orgname.platform.example.com`  so it just worked for us.

regardless happy to migrate off that spec.

is my summary of points above accurate?

OR13:
yep... and to be clear, did web supports `orgname.platform.example.com` and `platform.example.com/orgname`.

mkhraisha (approved):
lgtm

mprorock (approved):
Think we need at minimum did:web, with extended support for did-configuration

PR #50, opened 2022-01-18
doc: disaster recovery best practices

mprorock:


docs/index.html
```diff
@@ -480,6 +480,37 @@ <h2>Threat Model</h2>
...
+            <a href="https://nvlpubs.nist.gov/nistpubs/cswp/NIST.CSWP.10012020-draft.pdf">
+              Securing Data Integrity Against Ransomeware Attacks
+            </a>
+          </li>
+          <li>
+            <a href="https://csrc.nist.gov/publications/detail/sp/1800-11/final">
+              NIST SP 1800-11 - Data Integrity: Recovering from Ransomware and Other Destructive Events
+            </a>
+          </li>
+        </ul>
```
mkhraisha, line 507: Probably need another publication on recovery from outages or something that isn't explicitly an external attack, something maybe like this where its more general principles? https://docs.aws.amazon.com/whitepapers/latest/disaster-recovery-workloads-on-aws/disaster-recovery-options-in-the-cloud.html , although this is AWS only so a different source might be better.

mkhraisha (approved):

docs/index.html
OR13: https://github.com/w3c-ccg/traceability-interop/issues/52

OR13 (approved):

Issue #44, opened 2022-01-14
CI rule disable


mkhraisha:
Disabled a number of CI rules in PR #42  for the tests packages, as discussed on the PR @brownoxford  volunteered to fix those.

brownoxford:
I will be able to work on these issues this week.

PR #55, opened 2022-01-19
Refine proposal for discovery

OR13:
This PR removes the need to make additional network requests to discover api capabilities associated with an organization that supports the VC API.

Implementers would likely dynamically generate this `did.json` file based on the latest preferences the organization has for DIDs and their latest values... even with a DID URL present here, verification will fail if the URL is not listed in the corresponding did document.... in other words, planning for rotations in a did or even rotations of dids is a consideration when implementing this proposal.

We can consider giving the `didDocument.type` a value which makes it clear its purpose is to enable API discoverability.

PR #58, opened 2022-01-24
fix: address eslint issues in postman tests

brownoxford:
Address disabled eslint rules from #44.

Note that the `eslint-disable-line import/extensions` exception when importing the sha256 module is required because the default specifier resolution of "explicit" will not locate the module without the `.js` extension. An alternative would be to require `--experimental-specifier-resolution=node` on the command line when running the tests, but that seems like an unnecessary barrier to testing.

### DIFF
--- a/packages/tests-postman/README.md
+++ b/packages/tests-postman/README.md
@@ -16,7 +16,7 @@ All APIs and configs will be tested with Newman/Postman, and additional settings
 
 Install package with:
 ```bash
-npm i
+npm ci
 ```
 
 This package may be executed with the following:
@@ -105,9 +105,9 @@ JSON output is available for all tests and may be combined for analytics individ
 ```bash
 ├── bin - binary folder for cli test execution and integration into CI
 ├── collections - postman collections
-│   └── reference - postman examples related to this test suite
+│   └── reference - postman examples related to this test suite
 └── data - data packages for use with the postman collections
-    └── reference - examples
+    └── reference - examples
 ```
 
 The basic test flow is outlined in the diagram below:
--- a/packages/tests-postman/index.js
+++ b/packages/tests-postman/index.js
@@ -5,6 +5,9 @@ import newman from 'newman';
 import isUrl from 'is-url-superb';
 import ky from 'ky-universal';
 import fs from 'fs';
+
+// Default specifier resolution of "explicit" will not properly locate the
+// SHA 256 module unless `.js` extension is specified here.
 import sha256 from 'crypto-js/sha256.js'; // eslint-disable-line import/extensions
 
 console.log('Traceability Interop Testing');
@@ -24,7 +27,6 @@ program
     './collections/reference-credentials.json'
   );
 
-// disable-eslint-next-line program
 //     .option('-s, --service <file>', 'use the specified interop test collection', './collections/interop-credentials.json');
 
 program
@@ -38,8 +40,8 @@ program
     'use the specified reference VC data collection',
     './data/reference-credentials.json'
   );
-// program
-//     .option('-sd, --servicedata <file>', 'use the specified interop data collection', './data/interop-credentials.json');
+
+  //     .option('-sd, --servicedata <file>', 'use the specified interop data collection', './data/interop-credentials.json');
 
 program.option('-rd, --reportdir <folder>', 'use the specified service provider data collection', './newman');
 program.addOption(
@@ -74,6 +76,10 @@ const serviceData = JSON.parse(fs.readFileSync(program.opts().servicedata, 'utf8
 const referenceCollection = JSON.parse(fs.readFileSync(program.opts().reference, 'utf8'));
 const referenceData = JSON.parse(fs.readFileSync(program.opts().referencedata, 'utf8'));
 
+function getServiceIdx(name) {
+  return serviceData.findIndex((obj) => obj.name === name);
+}
+
 /**
  * Convenience definition documents the contents of an OAuth2Config object
  * @typedef {Object} OAuth2Config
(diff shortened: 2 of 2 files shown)


packages/tests-postman/README.md
```diff
@@ -16,7 +16,7 @@ All APIs and configs will be tested with Newman/Postman, and additional settings
 
 Install package with:
 ```bash
-npm i
+npm ci
```
OR13, line 19: why the need for ci install here?

packages/tests-postman/index.js
```diff
@@ -97,18 +103,18 @@ const referenceData = JSON.parse(fs.readFileSync(program.opts().referencedata, '
...
+  const { access_token_url: tokenURL, client_id: clientID } = serv.oauth2 || {};
+  if (typeof tokenURL === 'undefined') {
     throw Error(`"${name}" oauth2 configuration is missing "access_token_url" value`);
   }
-  if (typeof client_id === 'undefined') { // eslint-disable-line camelcase
+  if (typeof clientID === 'undefined') {
     throw Error(`"${name}" oauth2 configuration is missing "client_id" value`);
   }
-  const key = sha256(client_id);
+  const key = sha256(clientID);
```
OR13, line 117: clientID is public... I would prefer not to add hashing on top of identifiers that are expected to be public.
https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/

> The client_id is a public identifier for apps. Even though it’s public, it’s best that it isn’t guessable by third parties, so many implementations use something like a 32-character hex string. If the client ID is guessable, it makes it slightly easier to craft phishing attacks against arbitrary applications. It must also be unique across all clients that the authorization server handles.

Seems sufficient to just store `CLIENT_ID` as a secret in CI... no need to hash it.

I would expect to be able to trivially set these values from bash without running JS code.

packages/tests-postman/index.js
```diff
@@ -171,31 +170,43 @@ const urlExist = async (url) => {
...
+  const { names } = program.opts();
+  const services = serviceData.filter((serv) => names.includes('all') || names.includes(serv.name));
+
+  const results = [];
+
+  for (const serv of services) {
+    console.log('Checking', serv.name);
+
+    const urlSVC = serv.serviceProvider.provider.url;
+    const urlDID = `https://${serv.serviceProvider.baseURL}/.well-known/did-configuration.json`;
```
OR13, line 183: I dream of a change to this in the future.

packages/tests-postman/index.js
OR13: I guess hidden in what I am saying is the assertion that folks passing these tests will not be publishing CLIENT_IDs.... anywhere... My understanding was that we would use CI secrets for runtime tests, and likely obtain these values directly from service providers in other cases.


Perhaps we can expand on these expectations in a call.

packages/tests-postman/README.md
brownoxford: A clean install here will throw an error if there are any discrepancies between the `package.json` and `package-lock.json` files instead of modifying them to resolve the issues before installing the packages, which seemed more appropriate for the testing environment here.

packages/tests-postman/index.js
brownoxford: The hashing here is not for security, rather it is to map the alphabet allowed for client IDs to a subset of the more restrictive alphabet allowed for GitHub secret names. This facilitates mapping of client IDs to the client secret stored in GitHub, e.g., `CLIENT_SECRET_<SHA_HASH>`. I chose client_id for this purpose because it was the only field in `service-providers.json` guaranteed to map one-to-one to the client secret.

packages/tests-postman/index.js
brownoxford: This. 100%.

packages/tests-postman/index.js
```diff
@@ -24,7 +27,6 @@ program
     './collections/reference-credentials.json'
   );
 
-// disable-eslint-next-line program
 //     .option('-s, --service <file>', 'use the specified interop test collection', './collections/interop-credentials.json');
```
mkhraisha, line 30: do we need to keep those lines in and commented?
If they're not being used we should just remove them.

mkhraisha (approved):
is there a specific reason we switched out from using async/await to .then?

packages/tests-postman/index.js
OR13: ugh.... ok... perhaps we can use a "string nickname" instead? I don't like the idea or computing a hash just to save a secret in github.

brownoxford:
> is there a specific reason we switched out from using async/await to .then?

These were refactored to comply with the new `eslint rules`. Using await inside of a loop violates [Disallow await inside of loops (no-await-in-loop)](https://eslint.org/docs/rules/no-await-in-loop)

packages/tests-postman/index.js
brownoxford: We could absolutely do that instead, were you thinking along the lines of adding `oauth2.client_secret_envvar` to `service-providers.json` and then using `process.env[<client_secret_envver>]` to read the client secret from the environment?

packages/tests-postman/index.js
```diff
@@ -128,7 +134,7 @@ function getOAuth2Config(name) {
  */
 async function getBearerToken(name) {
   const params = getOAuth2Config(name);
-  const res = await ky
+  return ky
```
mkhraisha, line 137: my await/async comment was more targetted at a line like this one, where it seems the change was made without any loops. @brownoxford

packages/tests-postman/index.js
mprorock: none from my side

packages/tests-postman/index.js
mprorock: see #55

packages/tests-postman/index.js
brownoxford: I will create a new ticket to discuss this

packages/tests-postman/index.js
brownoxford: These dead options have been removed.

PR #59, opened 2022-01-25
api.did.actor serviceprovider

nissimsan:


nissimsan:
No oauth needed for this, so there oauth2 element or repo secret

brownoxford:
> No oauth needed for this, so there oauth2 element or repo secret

Is this temporary, or permanent? I understood that we were moving towards making missing/failing OAuth be a test failure condition, but that doesn't have to be the case.

packages/tests-postman/data/service-providers.json
```diff
@@ -56,5 +56,22 @@
...
+                "name": "Transmute DID Actor API",
+                "url": "https://api.did.actor",
+                "contact": {
+                    "name": "Orie Steele",
+                    "email": "orie@transmute.industries"
+                },
+                "markets": [ "steel", "e-commerce" ]
+            },
+            "baseURL": "https://api.did.actor",
+            "vcPrefix": "next",
```
OR13, line 73: I don't think this is right.... probably should be `api`... but I would have to look at the urls and match them up with what is published

OR13, line 73:
```suggestion
            "vcPrefix": "api",
```

OR13: Pulled form the console logs... 

```
https://api.did.actor/api/credentials/issue
```

mkhraisha (approved):

mprorock (approved):

Issue #57, opened 2022-01-21
Need the ability to correlate related credentials / presentations


mprorock:
need some well defined correlation id or similar

OR13:
Also related to https://github.com/w3c-ccg/traceability-interop/issues/85

We should close this once we can link to the vocab definition for correlation id.

OR13:
close this as a dupe

mprorock:
dupe #85

Issue #56, opened 2022-01-21
Need to define how and what workflows are


mprorock:
Need a definition and core structure and description for the notion of workflows in the context of tracing an item or shipment

OR13:
Related to https://github.com/w3c-ccg/traceability-interop/issues/85

We should close this once we can link to the vocab defintion for workflows.

OR13:
close this as a dupe.

mprorock:
dupe #85

PR #60, opened 2022-01-25
doc: add definition for workflow and correlation id

mprorock:
First step at addressing #56 and #57
I expect these definitions to evolve, and likely shorten if we add a discrete section on workflows, which is likely needed as there is a lot of nuance there.  I have tried to keep things as simple and generic as possible to account for virtual and physical traceability, and to map terms back to common graph theory terms.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -372,8 +372,8 @@
           Such an identifier is usually created upon the creation of the first
           node to be associated with a PATH, or upon the creation of a PATH
           (in whole or in part) after the fact.  This CORRELATION ID is attached to a
-          VC or other credentials by use of a property in the "credentialSubject" 
-          with the property name "correlationId".
+          VC, VP, or other credentials by use of a property named "correlationId" 
+          in the credential or as the domain with a presentation.
           <aside class="example" title="CORRELATIONID">
             <p>
               Use of a CORRELATION ID in a VC
@@ -386,11 +386,36 @@
                 ],
                 "type": ["VerifiableCredential"],
                 "issuer": "did:example:Wz4eUg7SetGfaUVCn8U9d62oDYrUJLuUtcy619",
+                "correlationId": "d2dfc100-9384-49e4-b547-5e7d84ec5bf5",
                 "credentialSubject": {
                   "originator": "Some User or System",
                   "product": "Blueberries",
-                  "correlationId": "d2dfc100-9384-49e4-b547-5e7d84ec5bf5"
                 }...
+            </pre>
+          </aside>
+          <aside class="example" title="CORRELATIONID">
+            <p>
+              Use of a CORRELATION ID in a VP as the domain
+            <p>
+            <pre class="js">
+              {
+                "@context": [
+                  "https://www.w3.org/2018/credentials/v1",
+                  "https://www.w3.org/2018/credentials/examples/v1"
+                ],
+                "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
+                "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
+                "verifiableCredential": [{  }],
+                "proof": {
+                  "type": "RsaSignature2018",
+                  "created": "2018-09-14T21:19:10Z",
+                  "proofPurpose": "authentication",
+                  "verificationMethod": "did:example:ebfeb1f712ebc6f1c276e12ec21#keys-1",
+                  
+                  "challenge": "1f44d55f-f161-4938-a659-f8026467f126",
+                  "domain": "d2dfc100-9384-49e4-b547-5e7d84ec5bf5"
+                }
+              }
             </pre>
           </aside>
         </dd>
--- a/docs/index.html
+++ b/docs/index.html
@@ -341,6 +341,59 @@
           <dfn data-lt="URI|URIs">URI</dfn>
         </dt>
         <dd>An identifier as defined by [[RFC3986]].</dd>
+        <dt>
+          <dfn data-lt="WORKFLOW|WORKFLOWs">WORKFLOW</dfn>
+        </dt>
+        <dd>
+          A distinct path for a distinct set of items being traced that either is declared
+          in advance by a user or system, or a root path that has already occurred
+          (in whole or in part). A WORKFLOW might branch, combine with others, or
+          terminate prematurely. It is an abstraction that in a traceability
+          context is inherently directed, and has a declared origin node,
+          and desired end node, with each node being a change in state or location.
+          Properties may be attached to nodes via credentials and correlated to one
+          another (and as a result, to the set of items) through use of a distinct
+          WORKFLOW CORRELATION ID.
+
+          In a real world context, a workflow might be created to describe one party
+          requesting a set of goods from some other location.  The workflow can then
+          be used to group and associate the nodes as those goods move from the origination
+          point (a node) to the end destination.  Any change in state, location, or 
+          control (legal or physical responsibility of some kind for the goods) would be
+          described as a node represented by a VC which has the workflow CORRELATION ID
+          attached to it. 
+        </dd>
+        <dt>
+          <dfn data-lt="CORRELATION ID|CORRELATION IDs|CORRELATIONID|CORRELATIONIDs">CORRELATION ID</dfn>
+        </dt>
+        <dd>
+          A unique identifier assigned to a WORKFLOW complying with [[RFC4122]].
+          A CORRELATION ID MUST conform with UUID Version 4.  
+          Such an identifier is usually created upon the creation of the first
+          node to be associated with a workflow, or upon the creation of a workflow
+          (in whole or in part) after the fact.  This CORRELATION ID is attached to a
+          VC or other credentials by use of a property in the "credentialSubject" 
+          with the property name "correlationId".
+          <aside class="example" title="CORRELATIONID">
+            <p>
+              Use of a CORRELATION ID in a VC
+            <p>
+            <pre class="js">
+              {
+                "@context": [
+                  "https://www.w3.org/2018/credentials/v1",
+                  "https://www.w3.org/2018/credentials/examples/v1"
+                ],
+                "type": ["VerifiableCredential"],
+                "issuer": "did:example:Wz4eUg7SetGfaUVCn8U9d62oDYrUJLuUtcy619",
+                "credentialSubject": {
+                  "workFlowOriginator": "Some User or System",
+                  "product": "Blueberries",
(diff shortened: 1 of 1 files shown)


docs/index.html
```diff
@@ -341,6 +341,59 @@ <h2>Terminology</h2>
...
+                "@context": [
+                  "https://www.w3.org/2018/credentials/v1",
+                  "https://www.w3.org/2018/credentials/examples/v1"
+                ],
+                "type": ["VerifiableCredential"],
+                "issuer": "did:example:Wz4eUg7SetGfaUVCn8U9d62oDYrUJLuUtcy619",
+                "credentialSubject": {
+                  "workFlowOriginator": "Some User or System",
+                  "product": "Blueberries",
+                  "correlationId": "d2dfc100-9384-49e4-b547-5e7d84ec5bf5"
```
OR13, line 392: this could go outside of the subject... and probably should since its a correlation for the credential

mprorock:
Have made a naming adjustment to "path" to avoid confusion between business process workflows as distinct from a "workflow" that is really describing a path within a graph

OR13:
> The properties in this field create a greater risk of correlation when the same values are used across more than one session or domain and the value does not change. 

https://www.w3.org/TR/vc-data-model/#signature-based-correlation

OR13:
> aud MUST represent (i.e., identify) the intended audience of the verifiable presentation (i.e., the verifier intended by the presenting holder to receive and verify the verifiable presentation).

https://www.w3.org/TR/vc-data-model/#jwt-encoding

OR13:
See also:

https://www.w3.org/TR/vc-data-model/#identifier-based-correlation

the fact that VP.id and VC.id is optional is a major reason for the need to be clever about using domain.

As a read the JSON-LD... I think actually this is legal...

```json
{
    "type": "RsaSignature2018",
    "created": "2018-09-14T21:19:10Z",
    "proofPurpose": "authentication", || "assertionMethod" // VP or VC
    "verificationMethod": "did:example:ebfeb1f712ebc6f1c276e12ec21#keys-1",
    "challenge": "1f44d55f-f161...", // nonce / channel authenticator required in VPs optional in VCs
    "domain": "4jt78h47fh47", // correlationId
    "jws": "eyJhbG..."
  }
```

OR13:
We should avoid building any new context terms if we can be clever about using the existing VC Data Model.

OR13:
Another note regarding the 'vc-api" and workflow "name" vs "id"...

In https://github.com/w3c-ccg/vc-api/pull/255 

The make a distinction between "name" and "id"...

"name" seems to correspond to a "Abstract Workflow Definition" for example "Renew Drivers License"

Whereas "id" seems too correspond to a specific instance of that workflow, "Alice renews her DL at the DMV in Fajardo PR."

mprorock:
so @OR13 two questions:
1) no need in your mind to correlate credentials on their own and explicitly independent of a VP?
2) how do you know if the `domain` is being used as a correlation identifier or not if as a part of a VP?

mprorock:
other options include making the correlation id a bit more structured and descriptive than a UUID on its own, e.g. a v5 or related hash based approach, the inputs of which would need to be defined

OR13:
> no need in your mind to correlate credentials on their own and explicitly independent of a VP?

I think ideally we move these correlations to the VP layer, so we can support a wider range of VCs... but of course they are optional at the VC layer...if we make them mandatory at the VC layer, be prepared to overhaul all the issuer systems... seems expensive / unrealistic to get that change to happen.

> how do you know if the domain is being used as a correlation identifier or not if as a part of a VP?

for all VP, we know that domain is used to correlate... but domain might be reused across multiple presentations.
for all VP, we know that challenge is used to correlate... but we expect it to either be unique to defend against replays, or repeated when parties are already in an authenticated channel (and replay attacks are therefor not a concern).

mkhraisha (commented):
Few points about this PR, great start btw thanks @mprorock 

1. Not super happy with the term path, process/workflow fit better IMO. would love some brainstorming on this on a call.
2. Do we need to introduce the term "correlation ID"? why are we not just calling it a path id?
3. Happy to use UUID v4.
4. I think the path id needs to be an array though, to allow for a VC to belong to multiple different paths.

## Scenario 

For the following assume the following scenario:
  1. Party A sends 3 VCs associated together to Party B.
  2. Party B wishes to create a two new VCs, a transformation event and a new product VC, and submit it to party C.
    * The transformation event would contain information that says product A is transformed by adding additive D to create Product B. (think smelting in steel or diluting in oil etc.)
  3. Party C wishes to verify all VCs are valid and relate to the same E2E flow.
  
## Assuming a path id may exist on both VCs and VPs

* We will have issues reconciling which identifier to use, which may cause developer friction. We can get around this friction by providing explicit purposes for the different pathIds for example VC path id are for a specific "sub-path" and terminate on a transformation and a VP path are for E2E path
* if we adopt the model of VC path id is for sub path and VP path id is for an E2E path, you can present VPs that contain VCs that belong to multiple different paths. 

  In the scenario outlined above this means that: 

  1. Party A creates the initial 3 VCs and associates them with VC path id 1, and submits a VP with pathId X to Party B
  2. Party B creates a new VC for a transformation event that captures the addition product M and A, and creates a new VC that represents product B with path id 2, and presents a VP to party C with pathId X that contains VCs from path id 1 and path id 2. This allows us to correlate the VCs to the same E2E flow, this also if we add an existing VC to the path id 2, we can correlate it to the same E2E flow.
  3. Party C can verify that all VCs are valid and relate to the same E2E flow.
  


## Assuming a path ID may only exist on VPs but not on VCs
   * What prevents you from presenting the same path to different parties with a different id? (in my mind nothing, which is a privacy advantage)
   * How does party B know what are all the VCs it must submit to party C? does it need to store the connection between the VCs received from party A? 
   * How do VCs added from other workflows carry their history? is this even relevant?
  
   
## Assuming a path ID may only exist on VCs but not on VPs
* How do we handle a VC that belongs to two separate paths? assume it is the end of one path and then is joined to a separate path. We would need to update the VC with a reference to a new path ( which is obviously not doable). 
* we COULD point to the prior VC from the current VC, similar to how git handles pointers, which allows us to have all the complexities of branching, the obvious cost however is correlation between VCs. Considering that these VCs are intended to be sent forward as a bundle however this might not be too big of an issue. It will still be a problem if you wish to share VCs X,Z but not Y with a party though.

OR13:
> I think the path id needs to be an array though, to allow for a VC to belong to multiple different paths.

I think thats not true... I think you only need to point to a parent `correlationId` and submit the data... you don't want to preclude tree and graph structures... including cycles, etc.

mkhraisha:
my point about `correlationId` being an array was under the assumption there is no pointer to a parent. if there is a pointer to a parent then I agree wholeheartedly

nissimsan:
It is not uncommon for a trade document to carry multiple references: [shipper's reference](https://github.com/w3c-ccg/traceability-interop/pull/60), [consignee's reference](https://service.unece.org/trade/uncefact/vocabulary/uncl1153/#Consignee's_reference), etc. 

It's unclear to me, could a parent-relationship solve for that? Could you guys add an example, pls?

OR13:
@nissimsan provide an example where 3 credentials all refer to the same correlationId.

nissimsan:
> @nissimsan provide an example where 3 credentials all refer to the same correlationId.

@OR13, that's sort of the inverse of what I'm asking. But let me describe a real-life scenario, which we would need to support: 

- A customer places an order to a Vendor for a container's worth of goods. She provides a PO# (say, `PO-end_cust123`).
- The Vendor order the production of these goods from the manufacturer. He provides both the his own internal PO# (`PO-vendor123`) as well as the customer's PO# (`PO-end_cust123`) to the manufacturer. 
- The manufacturer produces the goods ships directly to the end customer, therefore provides (`PO-end_cust123`) on the Packing List. 
- The manufacturer (consignor) also requests the carrier to add `PO-end_cust123` as consignee's reference on the BOL. 
- (The manufacturer's invoice to the vendor will only include `PO-vendor123`, no need for the customer PO# here).

OR13:
https://api.did.actor/workflow?name=linear-integration-example

Screen Shot 2022-01-27 at 9 44 01 AM

workflow (noun)
> the sequence of industrial, administrative, or other processes through which a piece of work passes from initiation to completion.

nissimsan:
I've worked on some examples for this discussion.

### Case
image
Based on this scenario, which is roughly what I described in my comment yesterday ^ (and scenario I PRed yesterday). I've excluded some less relevant document flows. Also, the order varies depending on payment terms, etc.

### Traditional 
image
Traditionally, this results in documentation like this. The “pouch” is what Customs would be handed (and also trade finance banks). The linkages are based on explicit document numbers (PL has a PO# field on it, BL has a booking# field on it, etc)

### Linear
image
Switching to the modern world, this is an example of linear document referencing. Both BL and PL mentions the container number, so when that has the same `type` and are given the same `id` , that establishes the correlation. 

### Branching
image
This is an example of a branching model. Customs still just get the documents in the pouch (policy requirements).

### Circular
image
And this is a circular example. The reference directions can also go around in circle, but I don't think that's terribly important point. 

### DID and Linked Data Potential
image
However, every time we define a `type` we establish a possible link across documents. So IMO we are already beyond what this discussion is about. These were some (bi-directional) links which can be established alone from the schemas and term definitions which we have built already. [This](https://github.com/w3c-ccg/traceability-vocab/issues/230) is a more important discussion, enabling data receivers (notably CBP) to establish these graphs.

mkhraisha:
Split Workflow Example
workflow examples - split workflow
Merge Workflow Example
workflow examples - Merge workflow

E2E Workflow Example
workflow examples - E2E Flow

OR13:
@mkhraisha best example yet.

docs/index.html
```diff
@@ -341,6 +341,84 @@ <h2>Terminology</h2>
           <dfn data-lt="URI|URIs">URI</dfn>
         </dt>
         <dd>An identifier as defined by [[RFC3986]].</dd>
+        <dt>
+          <dfn data-lt="PATH|PATHs">PATH</dfn>
+        </dt>
+        <dd>
+          A distinct path for a distinct set of items being traced that either is declared
+          in advance by a user or system, or a root path that has already occurred
+          (in whole or in part). A PATH might branch, combine with others, or
```
OR13, line 350: > A PATH might branch, combine with others, or terminate prematurely.

I feel like this is conflating a graph / tree, with the actual concept of a path (as in shortest path) in such a strucure.

OR13 (approved):

OR13:
this language needs work, but I prefer to refine it, in a subsequent PR.

OR13:
We might consider making `presentation.id` a required attribute.

nissimsan (approved):

PR #65, opened 2022-01-26
Add revocation list endpoint

OR13:
in our platform we host these credential at a different endpoint:

```
/credentials/{guid}
```

However if would be easy for us to update these endpoints to match whatever we agree is required for revocation interop here.

As an issuer you do need to host these files somewhere, (or trust someone else to host them and update when their status changes).

This is to start that conversation, because changing a credential status does not actually work, unless you define how to get the credential revocation list.

I updated the PR to match what we have today.

### DIFF
--- a/docs/components/CredentialLinkedDataProof.yml
+++ b/docs/components/CredentialLinkedDataProof.yml
@@ -1,4 +1,4 @@
-ttile: Credential Linked Data Proof
+title: Credential Linked Data Proof
 type: object
 description: A JSON-LD Linked Data proof.
 properties:
--- a/docs/components/CredentialLinkedDataProof.yml
+++ b/docs/components/CredentialLinkedDataProof.yml
@@ -1,0 +1,29 @@
+ttile: Credential Linked Data Proof
+type: object
+description: A JSON-LD Linked Data proof.
+properties:
+  type:
+    type: string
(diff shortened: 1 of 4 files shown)


mprorock (approved):

docs/components/CredentialLinkedDataProof.yml
```diff
@@ -0,0 +1,29 @@
+ttile: Credential Linked Data Proof
```
brownoxford, line 1:
```suggestion
title: Credential Linked Data Proof
```

docs/issuer-service.yml
```diff
@@ -36,6 +36,25 @@ paths:
           description: invalid input!
         "500":
           description: error!
+
+  /credentials/{id}:
```
nissimsan, line 40: Is this the right name for it? Shouldn't it explicitly indicate revocation?

OR13: see our current api for why it is the way it is here.

OR13: We could of course change our API. 

But I would prefer to only do that if we need to based on what the spec here says.

OR13:
This PR has been over taken by https://github.com/w3c-ccg/traceability-interop/pull/69

PR #69, opened 2022-01-27
Fix OAS 3 spec def

OR13:
- open the door for JWT agility
- open the door for JsonWebSignature2020 agility
- define a resolve endpoint
- define discovery
- fix respec embedding.


brownoxford:
Can we add `.DS_Store` to `.gitignore` and remove it from this PR?

PR #70, opened 2022-01-31
feat: workflow presentation

nissimsan:
Data model for the workflow/correlation requirements. 

I based this primarily on: 
https://github.com/w3c-ccg/traceability-interop/pull/60
https://github.com/w3c-ccg/traceability-vocab/issues/292

`TraceablePresentation` could extend `VerifiablePresentation`. For example if we want it to enforce a mandatory presentation.id. But otherwise I prefer to keep them separate, so the holder choose specifically indicates the dual intention: 
```
    "type": 
      [
        "VerifiablePresentation",
        "TraceablePresentation"
      ],
```

### DIFF
--- a/docs/components/TraceablePresentation.yml
+++ b/docs/components/TraceablePresentation.yml
@@ -1,0 +1,96 @@
+title: TraceablePresentation
+type: object
+description: A JSON-LD Presentation which establishes traceability linkage to a workflow instance and workflow type. 
+  - type: object
+    properties:
+      workflow:
+        $ref: "./Workflow.yml"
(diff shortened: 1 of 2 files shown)


OR13 (approved):

mprorock (approved):

OR13:
I think this PR is on the wrong repo, should be opened on trace vocab... you can't define a context bound term here.

OR13 (changes requested):
wrong repo

nissimsan:
Oh, that's a good point, @OR13. My thinking was "interop = VP, vocab = VC". I'll bring this over tomorrow.

Issue #61, opened 2022-01-25
Use hard-coded key instead of dynamic (via SHA2) for client id lookup


brownoxford:
@OR13 brought [up this issue](https://github.com/w3c-ccg/traceability-interop/pull/58#discussion_r791139831) in another ticket, and we discussed possibly adding a key to the `oauth2` section of the service provider JSON instead of dynamically coming up with a key using SHA256 and the client_id value.

PR #73, opened 2022-02-01
refactor: remove use of sha256

brownoxford:
Pre-define envvar name in service-providers.json instead of creating it dynamically from client_id with SHA256

FIX: #61


OR13 (approved):

packages/tests-postman/data/service-providers.json
```diff
@@ -42,6 +43,7 @@
         "oauth2": {
             "access_token_url": "https://transmute-industries.auth0.com/oauth/token",
             "client_id": "CGACwnX3KuIk9bfNgb7pSAd9SZxdCAEl",
+            "client_secret_envvar": "CLIENT_SECRET_b0643936e30843b622c78ad353da19123e4583817491fe05e2e9d39824983bee",
```
mkhraisha, line 46: probably a good idea to make this a named secret too?
rather than have it be a long string that will make secrets management super hard

brownoxford: This can be changed to anything you like - I left it set to the current secret value to prevent BC break.

brownoxford: I created #77 for this

mkhraisha: changed to approved now that issue is created

mkhraisha (approved):

mprorock (approved):
@OR13 this and pr #74 might take care of a few things you were looking at

Issue #68, opened 2022-01-26
Add CI to run postman tests


brownoxford:
We need to add workflow items to run the postman tests on push (secrets are not passed on PR from remote repos). This may require some additional work in getting the suite ready to run in this context. We should also look into how we can publish results of test runs for review (newman produces html artifacts).

mprorock:
@brownoxford any updates on this?

PR #74, opened 2022-02-01
Add CI to run postman tests

brownoxford:
Update `ci.yml` action to include postman tests. Note that this PR is contingent on #73 being merged first.

FIX: #68

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,3 +16,25 @@ jobs:
         run: npm ci
       - name: Lint
         run: npm run lint
+
+  postman:
+    runs-on: ubuntu-latest
+    # With the exception of GITHUB_TOKEN, secrets are not passed to the runner
+    # when a workflow is triggered from a forked repository.
+    if: github.repository == github.event.repository.full_name
+    steps:
+      - name: Check out git repository
+        uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+      - name: Install dependencies
+        run: npm ci
+        working-directory: packages/tests-postman
+      - name: Run Postman tests
+        run: node .
+        working-directory: packages/tests-postman
+        env:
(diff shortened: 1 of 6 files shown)


brownoxford:
Note that the `if` condition for the new `postman` tests isn't quite working right as it ran for this PR. I'll create a separate issue for that.

mprorock (approved):

.github/workflows/ci.yml
```diff
@@ -15,4 +15,26 @@ jobs:
...
+          node-version: 16.x
+      - name: Install dependencies
+        run: npm ci
+        working-directory: packages/tests-postman
+      - name: Run Postman tests
+        run: node .
+        working-directory: packages/tests-postman
+        env:
+          CLIENT_SECRET_VC_MESUR_IO: ${{ secrets.CLIENT_SECRET_VC_MESUR_IO }}
+          CLIENT_SECRET_b0643936e30843b622c78ad353da19123e4583817491fe05e2e9d39824983bee: ${{ secrets.CLIENT_SECRET_b0643936e30843b622c78ad353da19123e4583817491fe05e2e9d39824983bee }}
```
mkhraisha, line 40: similar to my comment on the other PR, would much prefer this to be `CLIENT_SECRET_TRANSMUTE`

.github/workflows/ci.yml
```diff
@@ -15,4 +15,26 @@ jobs:
...
+        uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+      - name: Install dependencies
+        run: npm ci
+        working-directory: packages/tests-postman
+      - name: Run Postman tests
+        run: node .
```
mkhraisha, line 36: nit: this should be `npm run test:all`

also any specific reason we're cding into the packages file and not just running it from base with lerna?

brownoxford: Nope, just my lack of knowledge on using lerna :) Looking into fixing this now.

brownoxford: ok, updated scripts to use lerna.

OR13 (approved):

PR #84, opened 2022-02-07
Add revocation test

OR13:
This PR adds authenticated credential revocation tests

### DIFF
--- a/.github/workflows/regression-revocation.yml
+++ b/.github/workflows/regression-revocation.yml
@@ -35,50 +35,6 @@ jobs:
           token_endpoint: ${{secrets[matrix.token_endpoint]}}
           api_base_url: ${{secrets[matrix.api_base_url]}}
         run: |
-          npx newman run ./docs/postman/regression/revocation.collection.json \
-          --env-var ORGANIZATION_DID_WEB=$organization_did_web \
-          --env-var CLIENT_ID=$client_id \
-          --env-var CLIENT_SECRET=$client_secret \
-          --env-var TOKEN_AUDIENCE=$token_audience \
-          --env-var TOKEN_ENDPOINT=$token_endpoint \
-          --env-var API_BASE_URL=$api_base_url \
-          --reporters cli,jsonname: Revocation Tests
-
-on:
-  workflow_dispatch:
-
-jobs:
-  build:
-    runs-on: ubuntu-latest
-    strategy:
-      # See matrix of secrets...
-      # https://sbulav.github.io/terraform/github-actions-matrix-secrets/
-      matrix:
-        include:
-          - organization_did_web: TRANSMUTE_STAGING_ORGANIZATION_DID_WEB
-            client_id: TRANSMUTE_STAGING_CLIENT_ID
-            client_secret: TRANSMUTE_STAGING_CLIENT_SECRET
-            token_audience: TRANSMUTE_STAGING_TOKEN_AUDIENCE
-            token_endpoint: TRANSMUTE_STAGING_TOKEN_ENDPOINT
-            api_base_url: TRANSMUTE_STAGING_API_BASE_URL
-
-    steps:
-      - name: Begin CI...
-        uses: actions/checkout@v2
-      - name: Use Node 14
-        uses: actions/setup-node@v1
-        with:
-          node-version: 14.x
-
-      - name: Run Tests
-        env:
-          organization_did_web: ${{secrets[matrix.organization_did_web]}}
-          client_id: ${{secrets[matrix.client_id]}}
-          client_secret: ${{secrets[matrix.client_secret]}}
-          token_audience: ${{secrets[matrix.token_audience]}}
-          token_endpoint: ${{secrets[matrix.token_endpoint]}}
-          api_base_url: ${{secrets[matrix.api_base_url]}}
-        run: |
           npx newman run ./docs/tutorials/revocation/revocation.collection.json \
           --env-var ORGANIZATION_DID_WEB=$organization_did_web \
           --env-var CLIENT_ID=$client_id \
@@ -86,4 +42,4 @@ jobs:
           --env-var TOKEN_AUDIENCE=$token_audience \
           --env-var TOKEN_ENDPOINT=$token_endpoint \
           --env-var API_BASE_URL=$api_base_url \
-          --reporters cli,json
+          --reporters cli,jsonname: Revocation Tests
--- a/.github/workflows/regression-revocation.yml
+++ b/.github/workflows/regression-revocation.yml
@@ -1,0 +1,89 @@
+name: Revocation
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      # See matrix of secrets...
+      # https://sbulav.github.io/terraform/github-actions-matrix-secrets/
+      matrix:
+        include:
+          - organization_did_web: TRANSMUTE_STAGING_ORGANIZATION_DID_WEB
+            client_id: TRANSMUTE_STAGING_CLIENT_ID
+            client_secret: TRANSMUTE_STAGING_CLIENT_SECRET
+            token_audience: TRANSMUTE_STAGING_TOKEN_AUDIENCE
+            token_endpoint: TRANSMUTE_STAGING_TOKEN_ENDPOINT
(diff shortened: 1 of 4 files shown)


docs/tutorials/revocation/README.md
```diff
@@ -1,5 +1,6 @@
 In this tutorial, you will learn:
 
+1. obtain an issuer id for an organization identified via a did:web
 1. issue revocable credentials
 1. verify a credential has not been revoked
```
mkhraisha, line 5:
```suggestion
1. verify that a credential has not been revoked
```

.github/workflows/regression-revocation.yml
```diff
@@ -0,0 +1,89 @@
...
+            token_endpoint: TRANSMUTE_STAGING_TOKEN_ENDPOINT
+            api_base_url: TRANSMUTE_STAGING_API_BASE_URL
+
+    steps:
+      - name: Begin CI...
+        uses: actions/checkout@v2
+      - name: Use Node 14
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
```
mkhraisha, line 27:
```suggestion
          node-version: 16.x
```

.github/workflows/regression-revocation.yml
```diff
@@ -0,0 +1,89 @@
...
+            client_id: TRANSMUTE_STAGING_CLIENT_ID
+            client_secret: TRANSMUTE_STAGING_CLIENT_SECRET
+            token_audience: TRANSMUTE_STAGING_TOKEN_AUDIENCE
+            token_endpoint: TRANSMUTE_STAGING_TOKEN_ENDPOINT
+            api_base_url: TRANSMUTE_STAGING_API_BASE_URL
+
+    steps:
+      - name: Begin CI...
+        uses: actions/checkout@v2
+      - name: Use Node 14
```
mkhraisha, line 24:
```suggestion
      - name: Use Node 16
```

.github/workflows/regression-revocation.yml
```diff
@@ -0,0 +1,89 @@
...
+          npx newman run ./docs/postman/regression/revocation.collection.json \
+          --env-var ORGANIZATION_DID_WEB=$organization_did_web \
+          --env-var CLIENT_ID=$client_id \
+          --env-var CLIENT_SECRET=$client_secret \
+          --env-var TOKEN_AUDIENCE=$token_audience \
+          --env-var TOKEN_ENDPOINT=$token_endpoint \
+          --env-var API_BASE_URL=$api_base_url \
+          --reporters cli,jsonname: Revocation Tests
+
+on:
```
mkhraisha, line 47: this seems wrong?
probably should either be a job under the existing workflow OR in its own workflow file.

mkhraisha (changes requested):

.github/workflows/regression-revocation.yml
OR13: This is in its own workflow file.

I think we should avoid putting all the jobs in the same file, especially for smaller set of basic feature validation.

It's nice to be able to confirm a simple feature (like revocation) in isolation form a wider range of tests.

I think ultimately we will need both small atomic tests, and larger integration tests.

mprorock: yes, I think orie has the right of it - basically quick feature tests that can be run directly, but also get incorporated in as part of baseline capabilities testing, and as a larger part of the interop suite 

@brownoxford - thoughts? basically I think modular is good

mprorock (approved):

.github/workflows/regression-revocation.yml
mkhraisha: I agree it should be on its own workflow, but here it looks like you have two separate workflows in one file, one to eventually run: ` npx newman run ./docs/postman/regression/revocation.collection.json` and one to eventually run `npx newman run ./docs/tutorials/revocation/revocation.collection.json `
not sure why the second one isnt just another step under the initial workflow or just another job under it.

OR13: hmm, maybe i got the path wrong?

.github/workflows/regression-revocation.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Run Tests
+        env:
+          organization_did_web: ${{secrets[matrix.organization_did_web]}}
+          client_id: ${{secrets[matrix.client_id]}}
+          client_secret: ${{secrets[matrix.client_secret]}}
+          token_audience: ${{secrets[matrix.token_audience]}}
+          token_endpoint: ${{secrets[matrix.token_endpoint]}}
+          api_base_url: ${{secrets[matrix.api_base_url]}}
+        run: |
+          npx newman run ./docs/tutorials/revocation/revocation.collection.json \
```
OR13, line 82: `/docs/tutorials/revocation/revocation.collection.json` is the file commited in this PR.

.github/workflows/regression-revocation.yml
OR13: I think I got the paths right, but if not my intention was:

1. a folder per collection with readme instructions on how to run it, and how it was built and screenshots
2. a workflow file per collection that allows a vendor to test that collection in isolation

I think I sorta accomplished these goals, but its possible I was tired when I made the PR and the paths are wrong.


Its a bit anoying since you can't test a PR like this easily until after it is merged, due to the use of `on dispatch`.

mkhraisha: I think what im trying to say is that you're testing the same workflow twice.
currently the file reads like this:

```
Name: Revocation
on: workflow_dispatch:
jobs: DO FIRST JOB
on: workflow_dispatch:
jobs: DO SECOND JOB
```
and both jobs look identical, except for that path, so either you made a mistake and are running the same job twice or they are indeed two distinct paths you want to test and in that case the second job's last step can just be appended to the first job.
if its still not clear happy to do the changes and show you what i mean

OR13: ahh yes.... the path can probably be factored into a single workflow... I just didn't do that in this PR.

There are cases where we might need different env vars, but I think we should strive to eliminate that :)

happy to see this sorted out, in subsequent PR.

.github/workflows/regression-revocation.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Run Tests
+        env:
+          organization_did_web: ${{secrets[matrix.organization_did_web]}}
+          client_id: ${{secrets[matrix.client_id]}}
+          client_secret: ${{secrets[matrix.client_secret]}}
+          token_audience: ${{secrets[matrix.token_audience]}}
+          token_endpoint: ${{secrets[matrix.token_endpoint]}}
+          api_base_url: ${{secrets[matrix.api_base_url]}}
+        run: |
+          npx newman run ./docs/postman/regression/revocation.collection.json \
```
mkhraisha, line 38: This url doesn't actually exist, unless im completely mistaken.

I think you can literally just remove the first half of this workflow and make sure to update the second half to contain 16.x and this should be enough.

.github/workflows/regression-revocation.yml
brownoxford: > yes, I think orie has the right of it - basically quick feature tests that can be run directly, but also get incorporated in as part of baseline capabilities testing, and as a larger part of the interop suite
> 
> @brownoxford - thoughts? basically I think modular is good

I believe in separate units wherever possible, and we can aggregate them and pass data via environmental outputs.

OR13:
I am dumb, I copy pasted twice.

.github/workflows/regression-revocation.yml
OR13: There is another issue to track this.

OR13:
@mkhraisha should be good to rereview.

mkhraisha (approved):

mkhraisha:
awesome lgtm!

docs/tutorials/revocation/README.md
```diff
@@ -1,6 +1,7 @@
 In this tutorial, you will learn:
 
+1. obtain an issuer id for an organization identified via did:web
 1. issue revocable credentials
-1. verify a credential has not been revoked
+1. verify that a credential has not been revoked
 1. revoke a credential
 1. verify a credential has been revoked
```
TallTed, line 7:
```suggestion
1. verify that a credential has been revoked
```

PR #104, opened 2022-02-22
Add vc-jwt example

OR13:
We opted to unwind an ill-advised attempt at storing VC_JWT in a form other then the compact JWT representation.


mprorock (approved):
This is really helpful and gives us a good starting place to work from on this topic

nissimsan (approved):

PR #109, opened 2022-03-07
fix: openapi.json should be up to date with latest changes in yml

brownoxford:
The `openapi.json` had not been recently updated. I ran `npm run preserve` to remove references to Pet and bring the file up to date.

### DIFF
--- a/docs/openapi/openapi.json
+++ b/docs/openapi/openapi.json
@@ -125,7 +125,18 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/Pet"
+                  "type": "object",
+                  "properties": {
+                    "didDocument": {
+                      "type": "object"
+                    },
+                    "didResolutionMetadata": {
+                      "type": "object"
+                    },
+                    "didDocumentMetadata": {
+                      "type": "object"
+                    }
+                  }
                 }
               }
             }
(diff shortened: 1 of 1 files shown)


mkhraisha (approved):

docs/openapi/openapi.json
```diff
@@ -125,7 +125,18 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/Pet"
+                  "type": "object",
```
OR13, line 128: consider making this a schema... like: https://github.com/transmute-industries/api.did.actor/blob/main/public/spec/schemas/DidResolutionResponse.yml

OR13: hmm... also... we should not be editing JSON... we should be editing yaml and generating json.

OR13 (changes requested):
see comments

docs/openapi/openapi.json
brownoxford: This is the output of `npm run preserve`, I did not hand-edit this.

brownoxford: It looks like `swagger-cli bundle` optimizes by inlining refs that are only used once, which is the case for this DidResolutionResponse item.

brownoxford:
> see comments

Moving conversation regarding manual maintenance of `openapi.json` (#117) and the awkward ref handling therein (#116) to their own tickets.

mprorock (approved):

OR13 (approved):

Issue #110, opened 2022-03-07
Github workflows contain syntax error


brownoxford:
We have this construct in several of the tests:
```--reporters cli,jsonname: Mill Test Report Certificate Tests```
which results in the following error:
```newman: could not find "jsonname:" reporter```

Everything after `,json` should be removed.

PR #111, opened 2022-03-07
fix: clear syntax issue with newman CLI tests

brownoxford:
Fix issues with unkown reporter `jsonname` when running newman tests from the command-line for github workflows.

FIX: #110

### DIFF
--- a/.github/workflows/regression-agility.yml
+++ b/.github/workflows/regression-agility.yml
@@ -42,4 +42,4 @@ jobs:
           --env-var TOKEN_AUDIENCE=$token_audience \
           --env-var TOKEN_ENDPOINT=$token_endpoint \
           --env-var API_BASE_URL=$api_base_url \
-          --reporters cli,jsonname: Agility Tests
+          --reporters cli,json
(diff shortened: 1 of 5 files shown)


TallTed:
I can't help but think the segments now being trimmed should instead become inline comments, if possible.  Is this possible?

.github/workflows/regression-agility.yml
```diff
@@ -42,4 +42,4 @@ jobs:
           --env-var TOKEN_AUDIENCE=$token_audience \
           --env-var TOKEN_ENDPOINT=$token_endpoint \
           --env-var API_BASE_URL=$api_base_url \
-          --reporters cli,jsonname: Agility Tests
+          --reporters cli,json
```
mkhraisha, line 45: Shouldn't these names be preserved and passed in as an argument to the reporter as a title?

mkhraisha (changes requested):

.github/workflows/regression-agility.yml
OR13: IMO, no... we should just output JSON, and then sanitize and merge it, then build a report from that... we should not be doing what we are now.

brownoxford: I believe the reporters also include the collection name as part of the output as well, so this would be redundant.

mkhraisha: If the reporter adds the collection name, then feel free to remove it.

mkhraisha (approved):

OR13 (approved):

mprorock (approved):
approved with note to fix conflicts

nissimsan (approved):

PR #114, opened 2022-03-07
fix: remove hard-coded server name from test

brownoxford:
Replace hard-coded server name in DID resolution request with environment variable.

FIX: 113

### DIFF
--- a/docs/tutorials/revocation/revocation.collection.json
+++ b/docs/tutorials/revocation/revocation.collection.json
@@ -1,285 +1,313 @@
 {
-  "info": {
-    "_postman_id": "144b0fa8-1fb1-4a81-98a1-16c8d06c285f",
-    "name": "Revocation",
-    "schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
-  },
-  "item": [
-    {
-      "name": "Get Access Token",
-      "event": [
-        {
-          "listen": "test",
-          "script": {
-            "exec": ["pm.environment.set(\"access_token\", pm.response.json().access_token);"],
-            "type": "text/javascript"
-          }
-        }
-      ],
-      "request": {
-        "method": "POST",
-        "header": [
-          {
-            "key": "Accept",
-            "value": "application/json",
-            "type": "text"
-          }
-        ],
-        "body": {
-          "mode": "raw",
-          "raw": "{\n    \"audience\": \"{{TOKEN_AUDIENCE}}\",\n    \"client_id\": \"{{CLIENT_ID}}\",\n    \"client_secret\": \"{{CLIENT_SECRET}}\",\n    \"grant_type\": \"client_credentials\"\n}",
-          "options": {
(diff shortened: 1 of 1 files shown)


brownoxford:
Note that postman export caused quite a lot of whitespace changes here. This PR is best viewed while ignoring whitespace, which can be done by clicking the cog on the "files changed" tab and checking the "ignore whitespace" box.

docs/tutorials/revocation/revocation.collection.json
```diff
@@ -1,285 +1,313 @@
...
+				"method": "GET",
+				"header": [
+					{
+						"key": "Accept",
+						"value": "application/json",
+						"type": "text"
+					}
+				],
+				"url": {
+					"raw": "{{API_BASE_URL}}/identifiers/{{ORGANIZATION_DID_WEB}}",
```
mkhraisha, line 73: This looks like it includes the identifiers change too?

brownoxford: ah - good catch. I'll pull that for now, as it should be part of a different PR that updates this URL in all postman tests.

docs/tutorials/revocation/revocation.collection.json
```diff
@@ -70,12 +70,11 @@
 					}
 				],
 				"url": {
-					"raw": "{{API_BASE_URL}}/identifiers/{{ORGANIZATION_DID_WEB}}",
```
OR13: I think we want this... the change should go the other direction... fixing all "resolution endpoints" to look like 

```
"{{API_BASE_URL}}/identifiers/{{ORGANIZATION_DID_WEB}}",
```

OR13 (changes requested):
We want this fix to got the other way.

docs/tutorials/revocation/revocation.collection.json
```diff
@@ -1,285 +1,312 @@
...
+				"method": "GET",
+				"header": [
+					{
+						"key": "Accept",
+						"value": "application/json",
+						"type": "text"
+					}
+				],
+				"url": {
+					"raw": "{{API_BASE_URL}}/{{ORGANIZATION_DID_WEB}}",
```
OR13, line 73:
```suggestion
					"raw": "{{API_BASE_URL}}/identifiers/{{ORGANIZATION_DID_WEB}}",
```

OR13: See https://w3c-ccg.github.io/did-resolution/#example-0

docs/tutorials/revocation/revocation.collection.json
brownoxford: Agreed - my apologies for acting without thinking this through. I've reverted the change.

OR13 (approved):

mprorock (approved):
approved with note to fix conflicts

nissimsan (approved):

PR #123, opened 2022-03-10
docs: add tutorial and tests for /credentials/issue endpoint

brownoxford:
This PR adds a fully documented tutorial with postman tests and GitHub workflow for making a `/credentials/issue` request to issue a verifiable credential.

I have some concerns about duplicate content in multiple tutorials that we may want to discuss and possibly address by denormalizing a little and adding links to other documents, but for now it seems that adding all of the information is best.


OR13 (approved):
beautiful.

nissimsan (approved):

PR #148, opened 2022-04-02
Automate Report Generation

OR13:
This PR

- [x] generate html and json reports from postman and host on github pages
- [x] sanitize json from postman
- [x] build an index for the reports
- [ ] setup perma urls for reports with w3id - https://github.com/perma-id/w3id.org/pull/2621

https://w3c-ccg.github.io/traceability-interop/reports ->

https://w3id.org/traceability/interoperability/reports

See follow up issue: https://github.com/w3c-ccg/traceability-interop/issues/151

PR #152, opened 2022-04-05
Workflow instance join

nissimsan:
This is a three-org workflow which demonstrates joining two individual workflow instance ids. Two organizations first make individual presentations to a third org (customs). When the correlation is realized, the workflows instances are joined with `["instance_1", "instance_2"]`. 

Currently wired to run on three separate tenants on Transmute, confirmed here: 
https://github.com/w3c-ccg/traceability-interop/runs/5829975747?check_suite_focus=true

image


BenjaminMoe (approved):

.github/workflows/regression-workflow-instance-join.yml
```diff
@@ -0,0 +1,84 @@
...
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      # Matrix paths are independent, and failure for one provider should not
+      # halt the running jobs testing another provider.
+      fail-fast: false
+      # See matrix of secrets...
+      # https://sbulav.github.io/terraform/github-actions-matrix-secrets/
+      matrix:
+        include:
```
brownoxford, line 16: What does everyone think about using [Alice/Bob](https://en.wikipedia.org/wiki/Alice_and_Bob) actor naming? I've started with that in #148 (.github/workflows/report-exchange.yml) with the goal of updating that matrix to various different permutations of actors.

It probably also makes sense to use full secret sets (even with some value duplication) for all of these instead of sharing secrets between orgs (e.g., re-use of `TRANSMUTE_PRODUCTION_TOKEN_ENDPOINT` for orgs 1, 2, and 3) to facilitate actor switching via matrix.

OR13 (approved):

mprorock (approved):

PR #158, opened 2022-04-05
fix: removing testing harness

nissimsan:


nissimsan:
This is fixing a PR we just merged - I will take the liberty to clean up my own mess right away.

PR #161, opened 2022-04-06
Revocation test fix

nissimsan:
This grabs the vc.id which was missing for revocation. 
It also tests for errors pre and post revocation.


brownoxford:
@nissimsan can we remove this un-merged branch?

Issue #41, opened 2021-12-13
Better error handling


OR13:
When errors occur, there should be helpful messages returned over http.

There should be an error message structure.

The structure should be used when handling 200 responses where the verification failed

There should also be a 400 error message structure, for cases where the client violates the api definition.

There should be test cases for errors, and each variant of an error.

There should be specific cases for revocation, expiration, preactivation.

OR13:
We use a structure like this internally:

```
VerificationResult:
      title: Verification Check
      description: Verification result for a Verifiable Credential
      type: array
      items:
        type: object
        properties:
          status:
            type: string
            example: 'good'
          title:
            type: string
            example: 'Activation'
          description:
            type: string
            example: 'This credential activated a month ago'
```

Turning checks into readable sentences improves devx, and ui rendering consistency.

OR13:
We need better error handling, we can't tell why things fail.

See https://github.com/w3c-ccg/vp-request-spec/pull/12

nissimsan:
This is between the presenting parties: define happy path and sad path interactions. What are the 400 error we expect can occur? What are relevant error messages which will help resovle?

nissimsan:
I might pick this up next time when I have better bandwidth.

OR13:
awaiting https://github.com/w3c-ccg/vc-api/pull/258

nissimsan:
next step: port over VC API error schemas, add postman for testing

OR13:
Need to port over the schemas and write postman tests that produce errors for all documented error cases.

OR13:
We still need to port over some error examples, ideally covering verify errors such as revocation.

nissimsan:
Necessary for addressing this: 
- Make negative tests.
- Define error JSONs.

OR13:
Use english language to describe error cases, then ensure that there is a postman collection that asserts the json structure for each error condition.

mkhraisha:
Initial Error cases when issuing a credential:

requests:

- [ ] issuer that does not support suite required by options
- [ ] revocation type requested that is not supported by the issuer
- [ ] missing credential subject
- [ ] term not defined in context

responses:
- [ ] missing credential subject
- [ ] the credential subject is not an object or a string
- [ ] missing context
- [ ] missing proof
- [ ] missing credential status when the revocable credential was requested
- [ ] missing credential identifier when the revocable credential was requested
- [ ] signature type of proof does not match the requested suite
- [ ] issuance date is not present

[...]

PR #162, opened 2022-04-06
Revocation error test

nissimsan:
Checks for errors pre and post revocation. 

(Note that verifiable_credential_id is missing, fixed in https://github.com/w3c-ccg/traceability-interop/pull/159)


brownoxford (changes requested):
This change in the testing causes a drift from the defined spec, which indicates that a verification response includes only a boolean `verified` property.

https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/VerificationResult.yml

This is related to #41, and we should probably wait until that is resolved before we change the tests.

nissimsan:
[Here](https://github.com/w3c-ccg/traceability-interop/pull/168)'s an update to the response object. It should unblock this - ref our discussion yesterday, @mprorock , @brownoxford .

OR13:
conflicts exist

nissimsan:
Merging, ref agreement yesterday. It implements https://github.com/w3c-ccg/traceability-interop/pull/168

PR #167, opened 2022-04-07
chore: clean up

nissimsan:
I might be slightly aggressive here :) but I propose that we get rid of the workflow we don't use. The list is a bit messy. 
I'm not deleting the collections, so we can easily add them back in later.


mkhraisha:
There are conflicts on this one, but I agree

PR #168, opened 2022-04-07
feat: detailed verification response

nissimsan:


docs/openapi/schemas/VerificationResult.yml
```diff
@@ -1,6 +1,15 @@
 title: Verification Result
 type: object
 properties:
-  verified:
-    type: boolean
-additionalProperties: true
+  checks:
```
OR13, line 4: i think we agree to remove this in the vc api...

OR13 (approved):

Issue #188, opened 2022-04-08
Presentations prover should send VP in response body


brownoxford:
Right now, the API defines the responses for the `/presentations/prove` request as follows:

```
  responses:
    "201":
      description: Expected response to a valid request
      content:
        application/json:
          schema:
            $ref: "../schemas/SerializedVerifiableCredential.yml"
    default:
      $ref: "../responses/UnexpectedError.yml"
```

I believe this should reference `../schemas/SerializedVerifiablePresentation.yml`, not `../schemas/SerializedVerifiableCredential.yml`

@OR13 if you confirm I will fix immediately.

OR13:
You are correct.

TallTed:
I guess [this](https://github.com/w3c-ccg/traceability-interop/issues/188#issuecomment-1095182377) should have been an approving CODEOWNER review on [the PR](https://github.com/w3c-ccg/traceability-interop/pull/190)?

PR #190, opened 2022-04-08
fix: update referenced response schema for presentation/prove

brownoxford:
FIX: #188


TallTed:
Uh... CODEOWNER reviews requested sometime Friday (github's "humanized" datetime of "3 days ago" makes exact datetime hard to discover), no review received, and PR merged ~noon ("1 hour ago") Monday? 

I don't have an argument with the PR itself, but this feels like an unnecessary process violation.  We should not get into such a habit.

PR #207, opened 2022-04-21
feat: removal of /api from path

nissimsan:
`/api` seems unnecessary. It is also inconsistently implemented.


mprorock (approved):
yeah, this is easily handled by a prefix on the server - of course we will all have to update implementations, or configs, again

BenjaminMoe (approved):

OR13 (approved):

mkhraisha (approved):

Issue #222, opened 2022-05-11
Presentations Verify Test Env Variables are Mapped Wrong


BenjaminMoe:
https://github.com/w3c-ccg/traceability-interop/blob/main/.github/workflows/interoperability-report.yml#L292
```
        env:
          organization_did_web: ${{secrets[format('{0}_ORGANIZATION_DID_WEB', matrix.actor)]}}
          client_id: ${{secrets[format('{0}_CLIENT_ID', matrix.actor)]}}
          client_secret: ${{secrets[format('{0}_CLIENT_SECRET', matrix.actor)]}}
          token_audience: ${{secrets[format('{0}_TOKEN_AUDIENCE', matrix.actor)]}}
          token_endpoint: ${{secrets[format('{0}_TOKEN_ENDPOINT', matrix.actor)]}}
          api_base_url: ${{secrets[format('{0}_API_BASE_URL', matrix.actor)]}}
        run: |
          npx newman run ./docs/tutorials/presentations-verify/presentations-verify.postman_collection.json \
          --env-var ORGANIZATION_DID_WEB=$organization_did_web \
          --env-var CLIENT_ID=$client_id \
          --env-var CLIENT_SECRET=$client_secret \
          --env-var TOKEN_AUDIENCE=$token_audience \
          --env-var TOKEN_ENDPOINT=$token_endpoint \
          --env-var API_BASE_URL=$api_base_url \
```

Fails with:

```
→ Get Access Token (Issuer)
  POST {{ISSUER_TOKEN_ENDPOINT}} [errored]
     getaddrinfo ENOTFOUND {{issuer_token_endpoint}}
  2. must return `200 Success` status
  3. response body must include non-empty access_token
  4. response body must represent `Bearer` token
  5. returned token must expire in the future
  6. `access_token` persisted to collectionVariables

→ Get Access Token (Verifier)
  POST {{VERIFIER_TOKEN_ENDPOINT}} [errored]
     getaddrinfo ENOTFOUND {{verifier_token_endpoint}}
  8. must return `200 Success` status
  9. response body must include non-empty access_token
 10. response body must represent `Bearer` token
 11. returned token must expire in the future
 12. `access_token` persisted to collectionVariables

→ Get Organization DIDs (Issuer)
  GET {{ISSUER_API_BASE_URL}}/identifiers/{{ISSUER_ORGANIZATION_DID_WEB}} [errored]
     getaddrinfo ENOTFOUND {{issuer_api_base_url}}
[...]

PR #219, opened 2022-05-10
fix: migrate presentations verify tutorial

brownoxford:
Migrate the "Presentations Verify" to the issuer/verifier multi-provider format used by the "Presentations Exchange" tutorial.

FIXES: #216

PR #223, opened 2022-05-12
fix: update matrix and environment for interop-presentations-verify

brownoxford:
This regression was introduced by PR #219 which contained an incomplete fix of Issue #216. This PR modifies the `interoperability-report.yml` workflow to match the changes made to the `regression-presentation-verify.yml` workflow, namely the introduction of multi-provider functionality in the test matrix and environment variables.

FIXES: #222

### DIFF
--- a/.github/workflows/interoperability-report.yml
+++ b/.github/workflows/interoperability-report.yml
@@ -296,12 +296,24 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Mavennet"
-            actor: "MAVENNET_STAGING"
-          - name: "mesur.io"
-            actor: "MESUR_IO_PRODUCTION"
-          - name: "Transmute"
-            actor: "TRANSMUTE_PRODUCTION"
+          - name: "mesur.io-Transmute"
+            issuer: MESUR_IO_PRODUCTION
+            verifier: TRANSMUTE_PRODUCTION
+          - name: "mesur.io-Mavennet"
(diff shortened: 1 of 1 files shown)


.github/workflows/interoperability-report.yml
```diff
@@ -296,12 +296,24 @@ jobs:
...
       matrix:
         include:
-          - name: "Mavennet"
-            actor: "MAVENNET_STAGING"
-          - name: "mesur.io"
-            actor: "MESUR_IO_PRODUCTION"
-          - name: "Transmute"
-            actor: "TRANSMUTE_PRODUCTION"
+          - name: "mesur.io-Transmute"
+            issuer: MESUR_IO_PRODUCTION
```
OR13, line 300: I question why this change is happening, my understanding is that all actors are capable of all roles.

there are 3 roles, issuer, holder and verifier.

if you are presenting to a verifier, you are a holder... this change is confusing.

OR13: We chatted, i think its best we follow up the naming conversation on an issue... in some ways, the current names are actually best, but non the less they are likely to confuse folks expecting to see "holder" when considering exchanges.

OR13 (approved):
merge when ready

.github/workflows/interoperability-report.yml
OR13: https://github.com/w3c-ccg/traceability-interop/issues/225

Issue #224, opened 2022-05-12
Only publish interop reports from main branch


brownoxford:
If the `interoperability-report.yml` is run from a branch other than `main`, e.g., during the course of testing a PR, the final `interop-publish` step should be omitted. While this will prevent the resulting report from being viewed, it does allow the `interoperability-report.yml` workflow to be tested in production without overwriting or otherwise polluting the published test results.

PR #226, opened 2022-05-12
feat: only publish reports from main branch

brownoxford:
Only publish reports if the workflow was triggered from the main branch.

FIXES: #224

### DIFF
--- a/.github/workflows/interoperability-report.yml
+++ b/.github/workflows/interoperability-report.yml
@@ -499,9 +499,11 @@ jobs:
     name: Publish
     runs-on: ubuntu-latest
 
-    # Always run, regardless of whether jobs that this one depends on have
-    # succeeded or failed.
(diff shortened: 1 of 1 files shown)


mkhraisha (approved):

.github/workflows/interoperability-report.yml
```diff
@@ -499,9 +499,11 @@ jobs:
     name: Publish
     runs-on: ubuntu-latest
 
-    # Always run, regardless of whether jobs that this one depends on have
-    # succeeded or failed.
-    if: ${{ always() }}
+    # Publishing must only happen when triggerd on the 'main' branch to prevent
```
TallTed, line 502:
```suggestion
    # Publishing must only happen when triggered on the 'main' branch to prevent
```

nissimsan (approved):

Issue #227, opened 2022-05-13
Credentials issue request should require credentials and proof type


brownoxford:
When making a request to the `/credentials/issue` endpoint, it seems like both the `credentials` and the `options` objects should be required, along with the `options.type` for specifying the signature type.

PR #228, opened 2022-05-13
fix: update "required" portions of credentials/issue request

brownoxford:
Update credentials/issue request to make "credential" and "options.type" required.

FIXES: #227

### DIFF
--- a/docs/openapi/resources/credential-issuer.yml
+++ b/docs/openapi/resources/credential-issuer.yml
@@ -13,6 +13,9 @@ post:
       application/json:
         schema:
           type: object
+          required:
+            - credential
+            - options
           properties:
             credential:
               $ref: '../schemas/Credential.yml'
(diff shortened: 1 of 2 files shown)


mprorock (approved):

docs/openapi/resources/credential-issuer.yml
```diff
@@ -13,6 +13,9 @@ post:
       application/json:
         schema:
           type: object
+          required:
+            - credential
+            - options
```
mkhraisha, line 18: isn't this supposed to be optional?

nissimsan (approved):

docs/openapi/resources/credential-issuer.yml
brownoxford: hmm. I may be misunderstanding this. What would the expected behavior be if a credential issue request is made without credentials to sign, or without specifying the signature type?

Issue #213, opened 2022-05-04
Create Presentation takes Crypto suite options via headers, Create Credentials via options


mkhraisha:
Currently Create Credentials takes a number of things via an options object including the crypto suite type:
![image](https://user-images.githubusercontent.com/19310771/166741162-c842043f-7cc7-4559-bada-b93f44a21a47.png)
 while the create presentations uses request headers:
 
![image](https://user-images.githubusercontent.com/19310771/166741422-8620bebc-ef30-4a41-b439-a157129b0498.png)

We should default to one or the other.

PR #230, opened 2022-05-16
moves type to options object

mkhraisha:
#213

Standardizes to Issuance object ( similar to credential issuance) if we agree on the other approach I'll migrate credential issuance.

### DIFF
--- a/docs/openapi/resources/presentation-prover.yml
+++ b/docs/openapi/resources/presentation-prover.yml
@@ -7,8 +7,6 @@ post:
   security:
     - OAuth2:
         - 'prove:presentations'
-  parameters:
-    - $ref: "../parameters/header/vpFormat.yml"
   requestBody:
     description: Parameters for creating the presentation.
     content:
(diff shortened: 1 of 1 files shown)


nissimsan (approved):

OR13 (changes requested):
change the path to make it clear that these are not "header" params.

include examples / confirm it works with the JWT test suite.

docs/openapi/resources/presentation-prover.yml
```diff
@@ -28,7 +25,10 @@ post:
                 challenge:
                   type: string
                   example: "1f44d55f-f161-4938-a659-f8026467f126"
-
+                type:
+                  $ref: "../parameters/header/proofType.yml"
```
OR13, line 29: this is not a header param

docs/openapi/resources/presentation-prover.yml
```diff
@@ -28,7 +25,10 @@ post:
...
+                vpFormat:
+                  $ref: "../parameters/header/vpFormat.yml"
```
OR13, line 31: this is not a header param

OR13: I don't think we need this option.

mprorock (changes requested):
should not be header params - also should drop vp format

brownoxford:
No actions taken since last meeting

nissimsan:
@OR13 , @mprorock , would you mind re-visiting this now? You're change requests seem to be addressed.

OR13 (approved):

OR13:
imo,mergeable.

PR #232, opened 2022-05-19
feat: adding implied oauth exchange endpoint

nissimsan:
I suppose this is how we expect to discover each others oauth presentation endpoints.

### DIFF
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -128,6 +128,7 @@ https://platform.example/organization/123/presentations/prove
 https://platform.example/organization/123/presentations/verify
 https://platform.example/organization/123/presentations/available
 https://platform.example/organization/123/presentations/submissions
+https://platform.example/organization/123/presentations
 ```
 
 These endpoints require the application to be authenticated:


nissimsan:
Note: out in "the real world" this won't be needed. If you already share client id and secret, you would also share this endpoint. But it will be practical for us, allowing us to continue getting the presentation-endpoint via did:web. If others have an easier way to share the presentation-endpoint pls share.

mkhraisha (approved):

mprorock (approved):

docs/discovery.md
```diff
@@ -128,6 +128,7 @@ https://platform.example/organization/123/presentations/prove
 https://platform.example/organization/123/presentations/verify
 https://platform.example/organization/123/presentations/available
 https://platform.example/organization/123/presentations/submissions
+https://platform.example/organization/123/presentations
```
OR13, line 131: intention is to support OAuth on this and exchange VPs that don't contain a proof?

nissimsan: Correct, that is the intention fo this

PR #235, opened 2022-05-19
Remove Merkle Proof Selective Disclosure

OR13:
There are no standard implementations.

Adding excessive "early stage experimental stuff" harms the usability of the profile and make its harder to understand.

If we are not intending to test it, we ought not to include it.

We are currently not intending to test either BBS+ or Merkle Disclosure Proofs.

I suggest we remove both until we have added support for them and agreed to test them.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/traceability-interop/pull/235.html" title="Last updated on May 19, 2022, 5:55 PM UTC (3a7c36f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/traceability-interop/235/60d5df4...3a7c36f.html" title="Last updated on May 19, 2022, 5:55 PM UTC (3a7c36f)">Diff</a>

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -453,7 +453,6 @@
           exist for this particular feature:
           <ul>
             <li>BBS+</li>
-            <li>Merkle Tree Proofs</li>
           </ul>
         </p>
         <section class="normative">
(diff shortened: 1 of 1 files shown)


nissimsan (approved):

mkhraisha (changes requested):
+1 to removing both, can you edit this PR to remove the BBS+ section that is just above it too?

mprorock (commented):
This is a developer note that makes clear that in theory this is something that "could" be done, but you shouldn't.  I am ok with leaving it in, but will not object to removing the section on merkle proofs

mkhraisha (approved):

PR #239, opened 2022-05-24
`TraceabilityAPI` Term Definition

nissimsan:
Removing the todo of https://github.com/w3c-ccg/traceability-vocab/pull/412. Merge the trace-vocab PR first.

### DIFF
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -34,11 +34,11 @@ HTTP GET https://platform.example/organization/123/did.json => didDocument.json
 
 2. Review the `serviceEndpoint`, `assertionMethod`, and `authentication` sections of the DID document.
 
-#### `services`
+#### `service`
 
 > Means of communicating or interacting with the DID subject or associated entities via one or more service endpoints. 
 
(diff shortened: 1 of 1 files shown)


mprorock (approved):

OR13 (changes requested):
Provide a DID Document example.

nissimsan:
I've updated the DID Document example with the necessary context, pointing at the new term we just defined.

```json
  "@context": [
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/traceability/v1"
  ],
```

With this `TraceabilityAPI` gets the proper term: 
`<did:web:platform.example:organization:123#traceability-api> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/traceability#TraceabilityAPI> .`

And yes, https://w3id.org/traceability#TraceabilityAPI resolves beautifully to reveal what it is we are doing here.

OR13 (approved):

BenjaminMoe (approved):

PR #241, opened 2022-06-01
OAuth-based presentation test

nissimsan:
Basic test of oauth-based presentations: allows presentation with a bearer token, asserts fail without it.


mkhraisha (approved):

nissimsan:
We'll take out the sad-path (4th test) from this - all negative cases should be bundled in conformance test suite.

mkhraisha (approved):

PR #253, opened 2022-06-06
Conformance testing

brownoxford:
Work-in-progress for conformance testing example of `/credentials/issue` endpoint.

Implemented:
* Happy path testing
* Multiple negative test cases that mutate request and expect 400 response
* Response validation against schema from this repository
* Script to convert yaml schema to json schema and inject into Postman tests

Still to do:
* Optional request element testing
* Authentication testing
* Scope checking
* Error response inspection and testing
* Automate yaml to json schema conversion and injection
* Add workflow to run test automatically

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -13,550 +13,1441 @@
 					"name": "Negative Testing",
 					"item": [
 						{
-							"name": "credential.@context:missing",
-							"event": [
+							"name": "Bad Request",
+							"item": [
 								{
-									"listen": "test",
-									"script": {
-										"exec": [
-											"pm.test(\"status code is 400\", function () {",
-											" pm.response.to.have.status(400);",
-											"});",
-											"",
-											"pm.test(\"response validates against schema\", function() {",
-											" const schemaString = pm.collectionVariables.get(\"responseSchema400\");",
-											" pm.response.to.have.jsonSchema(JSON.parse(schemaString));",
-											"});",
-											"",
-											""
+									"name": "credential:missing",
+									"event": [
+										{
+											"listen": "test",
+											"script": {
+												"exec": [
+													"pm.test(\"status code is 400\", function () {",
+													" pm.response.to.have.status(400);",
+													"});",
+													"",
+													"pm.test(\"response validates against schema\", function() {",
+													" const schemaString = pm.collectionVariables.get(\"responseSchema400\");",
+													" pm.response.to.have.jsonSchema(JSON.parse(schemaString));",
+													"});",
+													"",
+													""
+												],
+												"type": "text/javascript"
+											}
+										},
+										{
+											"listen": "prerequest",
+											"script": {
+												"exec": [
+													"let rawBody = pm.variables.get(\"rawBody\");",
+													"",
+													"// credential is required",
+													"delete rawBody.credential",
+													"",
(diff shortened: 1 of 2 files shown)


nissimsan:
Hi @brownoxford,

That's correct, the `client_credentials` flow does not allow for requesting a limited set of scopes for a grant. 

I'm testing this by making dedicated applications with varying scopes (at App-level, not Grant-level). That seems like a slightly less hackier approach. The problem is then we need to agree on a certain set of application definitions. Definitely cumbersome. Or we need to agree on a common way of creating Applications which is definitely scope creep and not the kind of thing we want to focus on. 

Can you describe more about how you can code around this on Auth0, pls? It might be the least bad option. 

A fourth option is to not test this section of the spec, I suppose. 

I suggest we bring this up on the call later today and discuss options. 

As a side note, it's confusing that OpenAPI makes it look like you can request scopes. I raised [this](https://github.com/OAI/OpenAPI-Specification/issues/2932) for insights. Seems like a mistake.

brownoxford:
I think OpenAPI is correct here, sort of. [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) indicates that client credentials grant access token requests may contain an _optional_ scope request. [Section 3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) of the same RFC states the following regarding scope requests:

> The authorization server MAY fully or partially ignore the scope
   requested by the client, based on the authorization server policy or
   the resource owner's instructions.

In the case of Auth0, they appear to partially ignore the scope request. If you specify an invalid scope, you get an error message.
```
{
  error: "access_denied",
  error_description: "Client has not been granted scopes: xwrite:presentations"
}
```
If you specify anything else (valid scope, empty string, omit scope entirely) all scopes assigned to the client are included in the JWT.

The workaround that can be used in Auth0 is [based on this post](https://community.auth0.com/t/client-credentials-request-ignores-scope-parameter/18496). Note that this code is for a Hook, not a rule or an action.

```
module.exports = function(client, scope, audience, context, cb) {
  var access_token = {};
  access_token.scope = scope;
  if (client.id === 'xxxxxxxx') {
    var requestedScopes = (context.body.scope || "").split(' ');
    access_token.scope = requestedScopes.filter(x => scope.indexOf(x) !== -1);
  }
  cb(null, access_token);
};
```

brownoxford:
@nissimsan I've updated the suite to include auth and scope checking so that we have an example of how that would work.

.github/workflows/conformance-test.yml
```diff
@@ -0,0 +1,51 @@
...
+      matrix:
+        include:
+          - name: "Mavennet"
+            actor: "MAVENNET_STAGING"
+          - name: "mesur.io"
+            actor: "MESUR_IO_PRODUCTION"
+          - name: "Transmute"
+            actor: "TRANSMUTE_PRODUCTION"
+    steps:
+      # Check out repo, setup node, and install dependencies.
```
TallTed, line 24:
```suggestion
      # Check out repo, set up node, and install dependencies.
```

TallTed (changes requested):

OR13:
I prefer not to have PRs sit this long with blocked status... how do we help move this along?

OR13 (approved):

BenjaminMoe (approved):

OR13:
Assuming we probably ok with CI failing for now.

PR #262, opened 2022-07-05
fix: use correct bearer token for verifier did resolution

brownoxford:
PR #260 updated the "Get Organization DIDs (Verifier)" request to use bearer auth, but used the `{{issuer_access_token}}` instead of the `{{verifier_access_token}}`.

This PR updates the request to use the proper token.

### DIFF
--- a/docs/tutorials/presentations-verify/presentations-verify.postman_collection.json
+++ b/docs/tutorials/presentations-verify/presentations-verify.postman_collection.json
@@ -1,8 +1,9 @@
 {
 	"info": {
-		"_postman_id": "c50743d4-76d4-4e3c-8256-ba0c5125492e",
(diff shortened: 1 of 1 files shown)


OR13 (approved):

OR13:
Does this "fix test failures" we are currently observing?

brownoxford:
The only difference I see when running the interop test in this branch is that the following test is green:
`Presentations Verify (mesur.io-Transmute, MESUR_IO_PRODUCTION, TRANSMUTE_PRODUCTION)`

OR13:
Cool, PRs that improve test performance should be tagged as such and celebrated as much as possible. Excellent work.

BenjaminMoe (approved):

PR #280, opened 2022-07-12
chore: fix add-mask usage

brownoxford:
add-mask needs to be called immediately


brownoxford:
merging security fix immediately

Issue #247, opened 2022-06-02
Define OAuth2 requirements for `/credentials/:credential-id`


brownoxford:
The goal of this issue is start discussion regarding the OAuth2 authentication requirements of the [`/credentials/:credential-id` endpoint](https://w3c-ccg.github.io/traceability-interop/openapi/#get-/credentials/-credential-id-).  This issue stems from the discussion on #187, and can be considered complete when the following items are resolved:

_**Please note**: the following lists are displayed as GitHub task lists for visual effect, not to suggest that there are actually additional tasks to be done._

The state of the checkboxes in this description represent the in-progress resolution of this issue. Until the `ready-for-close` label is applied, this issue should be considered in-progress. __This issue `MUST NOT` be closed until corresponding issues have been opened to implement the relevant conformance testing and, if necessary, modify the openapi spec.__

**Goal 1: Explicitly state OAuth2 requirements for the `/credentials/:credential-id` endpoint**
_One, and only one, of these options `MUST` be selected._

- [x] The `/credentials/:credential-id` endpoint MAY require OAuth2
- [ ] The `/credentials/:credential-id` endpoint MUST require OAuth2
- [ ] The `/credentials/:credential-id` endpoint MUST NOT require OAuth2

**Goal 2: Explicitly define scope requirements for this endpoint when authentication is in use:**
_One or more of these options `MUST` be selected **unless** the `MUST NOT` option was selected in **Goal 1**, in which case these options `MUST NOT` be selected. For requests that require OAuth2, each of the selected scopes `MUST` be present in an OAuth2 bearer token in order for the request to be permitted._

The following list of OAuth2 scopes is pulled from the current `openapi.yaml` file. If additional scopes are needed for the `/credentials/:credential-id` endpoint, they should be added here and an issue should be created to ensure that `openapi.yaml` is updated.
- [ ] `resolve:dids`
- [ ] `issue:credentials`
- [ ] `verify:credentials`
- [x] `read:credentials`
- [ ] `update:credentials`
- [ ] `prove:presentations`
- [ ] `verify:presentations`
- [ ] `submit:presentations`

brownoxford:
The `/credentials/:credential-id` endpoint MUST require OAuth2 to ensure that only authorized requests are allowed.

BenjaminMoe:
Require OAuth2 for credentials that are not public.
No authentication required for credentials that are public.

mkhraisha:
Who defines if a credential is public or not?

[...]

PR #286, opened 2022-07-14
Optional security for public credentials

nissimsan:
Ref https://github.com/w3c-ccg/traceability-interop/issues/247

I'm fairly sure this is the right syntax for optional security. The swagger editor accepts it too: 

image


nissimsan:
Replaced with a fresh PR

PR #297, opened 2022-07-14
Removing did oauth

nissimsan:


BenjaminMoe:
Is that the right end point?

https://github.com/w3c-ccg/traceability-interop/issues/251 (MUST)
```
  /identifiers/{did}:
    $ref: './resources/did.yml'
```

https://github.com/w3c-ccg/traceability-interop/issues/252 (MUST NOT)
```
  /did.json:
    $ref: './resources/api-configuration.yml'
```

PR #298, opened 2022-07-15
Remove security for did.json

nissimsan:
Getting rid of the lock: 
image

### DIFF
--- a/docs/openapi/resources/api-configuration.yml
+++ b/docs/openapi/resources/api-configuration.yml
@@ -1,10 +1,8 @@
 get:
   summary: API Configuration
   operationId: getTraceabilityAPIConfiguration
-
   tags:
     - Discovery
-  security: []
   responses:
     "200":
       description: Expected response to a valid request


BenjaminMoe (approved):

BenjaminMoe (approved):

mprorock (changes requested):
Did.json should be optional - it is a MAY

nissimsan:
@mprorock, no we landed on MUST NOT: 
https://github.com/w3c-ccg/traceability-interop/issues/252#issuecomment-1182184548

BenjaminMoe:
@mprorock Requiring OAuth2 on `/did.json` would make `did:web` unresolvable.

OR13 (changes requested):
This would break did:web.

OR13 (approved):
Ahh i read this wrong way, see the did web spec, this endpoint needs to be resolvable via GET, with no auth.

mprorock (approved):
revisted spec - lgtm

Issue #278, opened 2022-07-12
Interactive report graphs "don't look right"


brownoxford:
In the [conformance test output](https://w3c-ccg.github.io/traceability-interop/reports/conformance/), the two graphs in the result section don't appear to reflect the same values from the matrix above them. Need to investigate what is going on there and fix.

PR #300, opened 2022-07-15
chore: reporting updates and fixes

brownoxford:
* Update size of boxes in facet report to prevent overlapping
* Color pass/fail based on number of assertions passed or failed
* Clean up tooltips on facet report

FIX: #278

### DIFF
--- a/reporting/postman_reporter/report_charts.py
+++ b/reporting/postman_reporter/report_charts.py
@@ -188,12 +188,10 @@ def getFacet(df):
     df = df.groupby(["Provider", "Test Type", "Test Step"]).agg(["sum", "count"])
     df = df.set_axis(["passed", "tests"], axis="columns").reset_index()
 
-    print(df.head())
-
     # Map number of tests and passing assertions to a status.
     df.loc[df["passed"] == df["tests"], ["status"]] = "Pass"
-    df.loc[df["passed"] < df["tests"], ["status"]] = "Partial"
-    df.loc[df["passed"] == 0, ["tests"]] = "Fail"
+    df.loc[df["passed"] < df["tests"], ["status"]] = "Fail (Partial)"
+    df.loc[df["passed"] == 0, ["status"]] = "Fail"
 
     # Remove ephemeral columns
     df = df.drop(columns=["passed", "tests"])
@@ -259,7 +257,7 @@ def getSunburst(df, path=DEFAULT_REPORT_PATH):
 
     # Map number of tests and passing assertions to a status.
     df.loc[df["passed"] == df["tests"], ["status"]] = "Pass"
-    df.loc[df["passed"] < df["tests"], ["status"]] = "Partial"
+    df.loc[df["passed"] < df["tests"], ["status"]] = "Fail (Partial)"
     df.loc[df["passed"] == 0, ["status"]] = "Fail"
 
     # Remove ephemeral columns
@@ -293,7 +291,7 @@ def getTree(df, path=DEFAULT_REPORT_PATH):
 
     # Map number of tests and passing assertions to a status.
     df.loc[df["passed"] == df["tests"], ["status"]] = "Pass"
-    df.loc[df["passed"] < df["tests"], ["status"]] = "Partial"
+    df.loc[df["passed"] < df["tests"], ["status"]] = "Fail (Partial)"
     df.loc[df["passed"] == 0, ["status"]] = "Fail"
 
     # Remove ephemeral columns
--- a/reporting/postman_reporter/report_charts.py
+++ b/reporting/postman_reporter/report_charts.py
@@ -181,13 +181,36 @@ def getCards(df):
 
 
 def getFacet(df):
+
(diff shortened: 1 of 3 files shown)


reporting/postman_reporter/report_config.py
```diff
@@ -51,6 +51,11 @@
     "padding": "2rem 1rem",
 }
 
-COLOR_MAP = {"(?)": "darkred", "Pass": "darkgreen", "Fail": "darkred"}
+COLOR_MAP = {
+    "(?)": "darkgoldenrod",
+    "Pass": "darkgreen",
+    "Fail": "darkred",
+    "Partial": "darkgoldenrod",
```
OR13, line 58: hmm.... not sure about adding this.

OR13: Can you explain why its valuable?

OR13 (changes requested):
Unsure about adding "partial" to the test report. 

Need additional details to approve / deny.

defaulting to deny.

mprorock:
> Unsure about adding "partial" to the test report.
> 
> Need additional details to approve / deny.
> 
> defaulting to deny.

I definitely lean this way as well

reporting/postman_reporter/report_config.py
brownoxford: It serves to show when a test is partially passing, e.g., when a test case contains multiple assertions and only some of them fail. The other option would be to simply show those as failures.

If you would like to add a change request to modify `darkgoldenrod` to `darkred` it will produce the same output as before.

OR13: i see, I am ok with that, as long as we agree with respect to what "passing" means.

We should define passing as:

"All green, no yellow, and no red".

Anything other than "passing" is failing.

We are not planning on allowing for folks to say "we don't want to demonstrate interop on that feature, so its ok that we are "yelllow" for it...".

OR13 (changes requested):
change "Partial" to "Fail-partial", colors are good to keep as is.

reporting/postman_reporter/report_charts.py
```diff
@@ -181,13 +181,36 @@ def getCards(df):
...
+    # Group all of the assertions by step and provider so we can count how many
+    # assertions passed and failed for coloring purposes.
+    df = df.groupby(["Provider", "Test Type", "Test Step"]).agg(["sum", "count"])
+    df = df.set_axis(["passed", "tests"], axis="columns").reset_index()
+
+    print(df.head())
+
+    # Map number of tests and passing assertions to a status.
+    df.loc[df["passed"] == df["tests"], ["status"]] = "Pass"
+    df.loc[df["passed"] < df["tests"], ["status"]] = "Partial"
```
mprorock, line 195:
```suggestion
    df.loc[df["passed"] < df["tests"], ["status"]] = "Fail (Partial)"
```

reporting/postman_reporter/report_charts.py
```diff
@@ -225,35 +248,61 @@ def getFacet(df):
...
+    # Each row in the dataframe maps to a single assertion in the Postman tests.
+    # Group all of the assertions by step and provider so we can count how many
+    # assertions passed and failed for coloring purposes.
+    df = df[["Provider", "Test Type", "Test Step", "Passing"]].copy()
+    df = df.groupby(["Provider", "Test Type", "Test Step"]).agg(["sum", "count"])
+    df = df.set_axis(["passed", "tests"], axis="columns").reset_index()
+
+    # Map number of tests and passing assertions to a status.
+    df.loc[df["passed"] == df["tests"], ["status"]] = "Pass"
+    df.loc[df["passed"] < df["tests"], ["status"]] = "Partial"
```
mprorock, line 262:
```suggestion
    df.loc[df["passed"] < df["tests"], ["status"]] = "Fail (Partial)"
```

reporting/postman_reporter/report_charts.py
```diff
@@ -225,35 +248,61 @@ def getFacet(df):
...
```
mprorock, line 296:
```suggestion
    df.loc[df["passed"] < df["tests"], ["status"]] = "Fail (Partial)"
```

mprorock (changes requested):
see minor adjustments I suggested @brownoxford @OR13

brownoxford:
I modified the text used in the status column for partial failures to be `Fail (Partial)` and fixed a column name typo in `getFacet()`.

We can also play around with the text in the hover popup, e.g., stop showing `shape` and `size`, start showing `Passing: x/y`, etc.

OR13 (approved):

mprorock (approved):

brownoxford:
Here are some screenshots that show how reports will now look:

Screen Shot 2022-07-19 at 12 47 03 PM
Screen Shot 2022-07-19 at 12 47 30 PM
Screen Shot 2022-07-19 at 12 47 38 PM

OR13 (approved):

Issue #243, opened 2022-06-02
Define OAuth2 requirements for `/presentations/available`


brownoxford:
The goal of this issue is start discussion regarding the OAuth2 authentication requirements of the [`/presentations/available` endpoint](https://w3c-ccg.github.io/traceability-interop/openapi/#post-/presentations/available).  This issue stems from the discussion on #187, and can be considered complete when the following items are resolved:

_**Please note**: the following lists are displayed as GitHub task lists for visual effect, not to suggest that there are actually additional tasks to be done._

The state of the checkboxes in this description represent the in-progress resolution of this issue. Until the `ready-for-close` label is applied, this issue should be considered in-progress. __This issue `MUST NOT` be closed until corresponding issues have been opened to implement the relevant conformance testing and, if necessary, modify the openapi spec.__

**Goal 1: Explicitly state OAuth2 requirements for the `/presentations/available` endpoint**
_One, and only one, of these options `MUST` be selected._

- [ ] The `/presentations/available` endpoint MAY require OAuth2
- [ ] The `/presentations/available` endpoint MUST require OAuth2
- [x] The `/presentations/available` endpoint MUST NOT require OAuth2

**Goal 2: Explicitly define scope requirements for this endpoint when authentication is in use:**
_One or more of these options `MUST` be selected **unless** the `MUST NOT` option was selected in **Goal 1**, in which case these options `MUST NOT` be selected. For requests that require OAuth2, each of the selected scopes `MUST` be present in an OAuth2 bearer token in order for the request to be permitted._

The following list of OAuth2 scopes is pulled from the current `openapi.yaml` file. If additional scopes are needed for the `/presentations/available` endpoint, they should be added here and an issue should be created to ensure that `openapi.yaml` is updated.
- [ ] `resolve:dids`
- [ ] `issue:credentials`
- [ ] `verify:credentials`
- [ ] `read:credentials`
- [ ] `update:credentials`
- [ ] `prove:presentations`
- [ ] `verify:presentations`
- [ ] `submit:presentations`

BenjaminMoe:
**Goal 1:**

The `/presentations/available` endpoint MUST NOT require OAuth2

nissimsan:
> The /presentations/available endpoint MUST NOT require OAuth2

+1 on this. 

This is a DIDAuth endpoint. OAuth on top of it would be double-authentication.

brownoxford:
[...]

Issue #242, opened 2022-06-02
Define OAuth2 requirements for `/presentations/submissions`


brownoxford:
The goal of this issue is start discussion regarding the OAuth2 authentication requirements of the [`/presentations/submissions` endpoint](https://w3c-ccg.github.io/traceability-interop/openapi/#post-/presentations/submissions).  This issue stems from the discussion on #187, and can be considered complete when the following items are resolved:

_**Please note**: the following lists are displayed as GitHub task lists for visual effect, not to suggest that there are actually additional tasks to be done._

The state of the checkboxes in this description represent the in-progress resolution of this issue. Until the `ready-for-close` label is applied, this issue should be considered in-progress. __This issue `MUST NOT` be closed until corresponding issues have been opened to implement the relevant conformance testing and, if necessary, modify the openapi spec.__

**Goal 1: Explicitly state OAuth2 requirements for the `/presentations/submissions` endpoint**
_One, and only one, of these options `MUST` be selected._

- [ ] The `/presentations/submissions` endpoint MAY require OAuth2
- [ ] The `/presentations/submissions` endpoint MUST require OAuth2
- [x] The `/presentations/submissions` endpoint MUST NOT require OAuth2

**Goal 2: Explicitly define scope requirements for this endpoint when authentication is in use:**
_One or more of these options `MUST` be selected **unless** the `MUST NOT` option was selected in **Goal 1**, in which case these options `MUST NOT` be selected. For requests that require OAuth2, each of the selected scopes `MUST` be present in an OAuth2 bearer token in order for the request to be permitted._

The following list of OAuth2 scopes is pulled from the current `openapi.yaml` file. If additional scopes are needed for the `/presentations/submissions` endpoint, they should be added here and an issue should be created to ensure that `openapi.yaml` is updated.
- [ ] `resolve:dids`
- [ ] `issue:credentials`
- [ ] `verify:credentials`
- [ ] `read:credentials`
- [ ] `update:credentials`
- [ ] `prove:presentations`
- [ ] `verify:presentations`
- [ ] `submit:presentations`

BenjaminMoe:
**Goal 1**

The `/presentations/submissions` endpoint MUST NOT require OAuth2

TallTed:
> The `/presentations/available` endpoint MUST NOT require OAuth2

Justification? 

[...]

PR #317, opened 2022-07-22
Fix how endpoints without security is specified

nissimsan:
`/presentations/available` and `/presentations/submissions` are specified with `security: []`. While that does seem like a way to explicitly state that security is not allowed, the OAS doesn't seem to understand it, showing a lock for those endpoints: 
image

This PR removes `security: []`, which is also how we specified [getTraceabilityAPIConfiguration](https://w3c-ccg.github.io/traceability-interop/openapi/#get-/did.json).

### DIFF
--- a/docs/openapi/resources/presentation-submissions.yml
+++ b/docs/openapi/resources/presentation-submissions.yml
@@ -4,7 +4,6 @@ post:
   description: End a presentation exchange flow
   tags:
     - Presentations
-  security: []
   requestBody:
     description: A Verifiable Presentation to Store
     content:
--- a/docs/openapi/resources/presentation-available.yml
+++ b/docs/openapi/resources/presentation-available.yml
@@ -4,7 +4,6 @@ post:
   description: Start a presentation exchange flow
   tags:
     - Presentations
-  security: []
   requestBody:
     description: Description of the flow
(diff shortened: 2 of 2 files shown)


brownoxford (approved):

docs/openapi/resources/presentation-submissions.yml
```diff
@@ -4,7 +4,6 @@ post:
   description: End a presentation exchange flow
   tags:
     - Presentations
```
mprorock, line 6:
```suggestion
    - Presentations
        security:
            - OAuth2: [submit:presentations]
```

might be more inline with the spec?

mprorock (changes requested):

docs/openapi/resources/presentation-submissions.yml
brownoxford: @mprorock, `/presentations/available` and `/presentations/submission` MUST NOT require OAuth2, see #242 and #243.

mprorock: +1 switching to approved - my bad

mprorock (approved):

mkhraisha (approved):

PR #328, opened 2022-08-10
feat: add caching to token requests

brownoxford:
Each postman request now checks to see if the required OAuth2 token is available in a global token cache instead of making at least one, but sometimes two, token requests per Postman request.

This will help prevent overloading OAuth2 providers with requests an possibly running into 429 HTTP responses as we continue to add more tests to the conformance suite, or when a user is performing a lot of local testing.

The token cache is maintained as a Postman global variable, and may need to be cleared from time to time.

I would like to merge this out-of-band, please indicate whether you are ok with that when reviewing.


brownoxford:
Note - this also fixes an issue where negative testing of bad auth tokens was overwriting the auth token and causing 403 errors in downstream requests.

PR #366, opened 2022-08-25
chore: clarify request body for credentials status request

### LINKED ISSUE #358: Add more definition to credential status request body
Currently, the OpenAPI definition for the request body of `/credentials/status` is rather loose. This issue proposes that we restrict the definition a bit to make expectations clearer and testing easier.

[Current schema](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/resources/credential-status.yml):
```yaml
        schema:
          type: object
          description: Request for updating the status of an issued credential.
          properties:
            credentialId:
              type: string
            credentialStatus:
              type: array
              items:
                type: object
                properties:
                  type:
                    type: string
                  status:
                    type: string
```

Proposed Schema:

```yaml
        schema:
          type: object
          description: Request for updating the status of an issued credential.
          required:
            - credentialId
            - credentialStatus
          properties:
            credentialId:
              type: string
            credentialStatus:
              type: array
              maxItems: 1
              items:
                type: object
                required:
                  - type
                  - status
                properties:
                  type:
                    type: string
                    enum: ['RevocationList2020Status']
                  status:
                    description: Use a value of 0 (zero) to indicate "revoked", or 1 (one) to indicate not revoked.
                    type: integer
                    minimum: 0
                    maximum: 1
```

brownoxford:
This PR is an implementation of the proposed changes from #358, namely making the OpenAPI definition for `credentials/status` request bodies more strict.

fixes #358

### DIFF
--- a/docs/openapi/resources/credential-status.yml
+++ b/docs/openapi/resources/credential-status.yml
@@ -36,7 +36,7 @@ post:
                     # Traceability interoperability implementation must conform
                     # with VC-API, which uses string for the status property.
                     # https://w3c-ccg.github.io/vc-api/#update-status
-                    description: Use a string value of "0" (zero) to indicate "revoked", or "1" (one) to indicate not revoked.
+                    description: Use a string value of "1" (one) to indicate "revoked", or "0" (zero) to indicate not revoked.
                     type: string
                     enum: ['0', '1']
           example: {
--- a/docs/openapi/resources/credential-status.yml
+++ b/docs/openapi/resources/credential-status.yml
@@ -14,25 +14,35 @@ post:
         schema:
           type: object
           description: Request for updating the status of an issued credential.
+          required:
+            - credentialId
+            - credentialStatus
           properties:
-            # Only credentials with an id are revocable.
             credentialId:
               type: string
             credentialStatus:
               type: array
+              maxItems: 1
               items:
                 type: object
+                required:
+                  - type
+                  - status
                 properties:
                   type:
                     type: string
+                    enum: ['RevocationList2020Status']
                   status:
-                    type: string
+                    description: Use a value of 0 (zero) to indicate "revoked", or 1 (one) to indicate not revoked.
+                    type: integer
+                    minimum: 0
+                    maximum: 1
           example: {
               'credentialId': 'urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a',
               # // https://w3c-ccg.github.io/vc-status-rl-2020/
               # // If the binary value of the position in the list is 1 (one),
               # // the verifiable credential is revoked, if it is 0 (zero) it is not revoked.
-              'credentialStatus': [{ 'type': 'RevocationList2020Status', 'status': '0' }],
+              'credentialStatus': [{ 'type': 'RevocationList2020Status', 'status': 0 }],
             }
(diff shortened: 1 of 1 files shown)


docs/openapi/resources/credential-status.yml
```diff
@@ -14,25 +14,35 @@ post:
         schema:
           type: object
           description: Request for updating the status of an issued credential.
+          required:
+            - credentialId
```
nissimsan, line 18: This does't seem right. This should be just `id`, shouldn't it? 

https://www.w3.org/TR/vc-data-model/#identifiers

docs/openapi/resources/credential-status.yml
```diff
@@ -14,25 +14,35 @@ post:
...
                     type: string
+                    enum: ['RevocationList2020Status']
                   status:
-                    type: string
+                    description: Use a value of 0 (zero) to indicate "revoked", or 1 (one) to indicate not revoked.
+                    type: integer
+                    minimum: 0
+                    maximum: 1
           example: {
               'credentialId': 'urn:uuid:45a44711-e457-4fa8-9b89-69fe0287c86a',
```
nissimsan, line 41: Same here, obviously.

nissimsan (changes requested):

docs/openapi/resources/credential-status.yml
brownoxford: Hmm. I would argue that the request body here is _referencing_ a credential by ID, not _providing_ a credential and giving its ID (which would fall under the linked vc data model). If I understand things correctly, an `id` property generally serves to identify the object in which it occurs. When interpreted that way, I think that `credentialId` is more appropriate here.

OR13: This endpoint MUST align with https://w3c-ccg.github.io/vc-api/#update-status

OR13 (approved):

nissimsan (approved):

docs/openapi/resources/credential-status.yml
brownoxford: @OR13 in that case it sounds like I should modify the `type` property of each `credentialStatus` array item to be string with allowable values of "0" and "1" via `enum`, do you agree?

OR13: yes, pretty sure that endpoint has changed since it was created.

docs/openapi/resources/credential-status.yml
```diff
@@ -33,16 +33,18 @@ post:
...
                     enum: ['RevocationList2020Status']
                   status:
-                    description: Use a value of 0 (zero) to indicate "revoked", or 1 (one) to indicate not revoked.
-                    type: integer
-                    minimum: 0
-                    maximum: 1
+                    # Traceability interoperability implementation must conform
+                    # with VC-API, which uses string for the status property.
+                    # https://w3c-ccg.github.io/vc-api/#update-status
+                    description: Use a string value of "0" (zero) to indicate "revoked", or "1" (one) to indicate not revoked.
```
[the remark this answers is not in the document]
brownoxford: I believe you are correct on the swap - will fix.

nissimsan (approved):

PR #370, opened 2022-08-30
Many minor fixes add up to big fix of README.md

TallTed:


TallTed:
Something very similar to this should also be applied to traceability-vocab README.md

README.md
```diff
@@ -39,13 +39,13 @@ channel automatically.
 
 1. Select "Start Recording" to start audio recording
 
-1. Post the agenda link in the meeting chat box
+1. Post the agenda link (if non-standing) in the meeting chat box
```
brownoxford, line 42: I believe this should always be posted, as it triggers output on the HTML side when the meeting minutes are parsed. If we don't add this, no agenda will appear here: https://w3c-ccg.github.io/meetings/2022-08-23-traceability/

TallTed: OK. Then the standing agenda should be found at the URI in the instructions (which it is not, today). Better, an agenda should be sent out to the mailing list _before each and every meeting_, whether or not it's the "standing" agenda, and a link to *that message* should be posted to the chat at the start of the meeting.

README.md
```diff
@@ -39,13 +39,13 @@ channel automatically.
...
 
    `Agenda: https://github.com/w3c-ccg/traceability-interop/AGENDA.md`
```
brownoxford, line 44:
```suggestion
   `Agenda: https://github.com/w3c-ccg/traceability-interop/blob/main/AGENDA.md`
```

README.md
brownoxford: Agreed. I've added a suggested change to fix the URL for the existing process. I see what you mean regarding sending the agenda to the mailing list, and have [opened a separate issue](https://github.com/w3c-ccg/traceability-interop/issues/371) to address that.

TallTed:
```suggestion
1. Post the link to today's agenda in the meeting chat box. The link below leads 
   to the default "standing" agenda, and should be posted only if a date-specific 
   agenda was not sent to the mailing list.
```

TallTed:
@brownoxford -- Please re-review. I've made two adjustments based on your comments.

nissimsan (approved):

README.md
```diff
@@ -39,13 +39,15 @@ channel automatically.
 
 1. Select "Start Recording" to start audio recording
 
-1. Post the agenda link in the meeting chat box
+1. Post the link to today's agenda in the meeting chat box. The link below leads 
+   to the default "standing" agenda, and should be posted only if a date-specific 
+   agenda was not sent to the mailing list.
+
+   `Agenda: https://github.com/w3c-ccg/traceability-interop/blob/main/AGENDA.md`
```
brownoxford, line 46:
```suggestion
1. Post the link to today's agenda in the meeting chat box. The link below leads 
   to the default "standing" agenda. If a date-specific agenda was sent to the
   mailing list prior to the meeting, a link to that archived email should be used
   instead.

   `Agenda: https://github.com/w3c-ccg/traceability-interop/blob/main/AGENDA.md`
```
I agree with your suggestion that what we *really* should be doing here is posting the agenda to the mailing list prior to the meeting, and linking to that post _instead_ of using a standing agenda link. I have opened #371 to address that change separately.

I think it is important that we do not suggest that it is ok to omit this item from the meeting chat, as it would prevent the agenda link from showing in the rendered HTML minutes.

brownoxford (approved):

mkhraisha (approved):

OR13 (approved):

PR #378, opened 2022-09-06
tests: add explicit bad type checks

brownoxford:
This PR adds some explicit bad type checks for `credential.proof` in the request body for the `/credentials/verify` endpoint. Adding these checks to iterate over the possible bad types exposed some errors in our implementation, and should have been included in the testing from the start.

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -3599,6 +3599,311 @@
 										}
 									},
 									"response": []
+								},
+								{
+									"name": "credentials_verify:proof:string",
+									"event": [
+										{
+											"listen": "prerequest",
+											"script": {
+												"exec": [
+													"let rawBody = pm.variables.get(\"dummyVC\");",
+													"",
+													"// credential.proof is required to be an object for embedded proof VCs",
+													"rawBody.proof = \"string\";",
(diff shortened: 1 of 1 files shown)


nissimsan (approved):

nissimsan:
@brownoxford, wasn't there an issue for this which should be referenced?

brownoxford:
> @brownoxford, wasn't there an issue for this which should be referenced?

I don't believe so, this came up while I was working on our implementation and trying to get it to pass https://w3c-ccg.github.io/vc-api-verifier-test-suite/

Is it bad practice to create PRs without issues? I'm happy to stop doing that :)

nissimsan:
No no, not bad practice. (And if it was, who would I be to point out bad practices! xD ) Just wanted to make sure we link to issues if they do exist so we can get them closed.

TallTed:
Welll.... It *is* bad practice to make a PR without a preceding corresponding Issue _**if there is, or is likely to be, any controversy about the PR's proposed resolution of the underlying but unstated Issue.**_ Which doesn't seem to be the case about most if not all of your recent PRs, @brownoxford. 

All that said, creating an Issue before a fixing PR remains _generally_ best-practice.

PR #380, opened 2022-09-07
Conformance alignment to type array convention

nissimsan:
This changes the expected service type from `"type": "TraceabilityAPI"` to `"type": ["TraceabilityAPI"]`, aligning with the type array convention we've moving to on trace-vocab.


brownoxford (commented):
These embedded schemas should not be edited in the postman suite - they are automatically generated from the OpenAPI definitions by a GitHub workflow item using the `tests/update_conformance_schemas.sh` script.

Any desired changes to the OpenApi validation like this should be made to the OpenAPI spec itself.

brownoxford:
Note: This PR should also modify any examples containing service definitions, e.g., `discovery.md`.

nissimsan:
Thanks @brownoxford. Didn't know this is how the collection gets generated, I will update accordingly. 

Also thanks for pointing out the did core data model. My main concern is that an array certainly shouldn't fail; whether we disallow a string option is a separate discussion.

OR13 (approved):

OR13:
I suggest maybe deleting dicovery.md and updating the respec document...

mkhraisha (approved):

brownoxford (changes requested):
Just wanted to add a "request changes" comment, as the current changes will be overwritten next time the OpenAPI spec changes and these embedded values are regenerated.

nissimsan:
I now updated the OAS to take the string array. 
Also deleted discovery.md 🎉 and updated the respec. 

@brownoxford, can I ask you to take another look, pls? 
Note that there are a couple of the tutorial READMEs which might need to be updated too? I don't suppose they get generated. Wanted to talk it over with you before touching them.

brownoxford (approved):
Looks great, thanks for updating!

nissimsan:
I will look into how to resolve this, and merge once resolved

Issue #381, opened 2022-09-07
Drop responseSchema4xx response body `code`


nissimsan:
I don't see the point of including `code` in the error response bodies. I suggest we get get rid of it (on the OAS and conformance tests). 

```
{
"code": 0,   // <- this
"message": "Bad Request: Your request body does not conform to the required schema",
"details": "string"
}
```

nissimsan:
Ping @brownoxford

nissimsan:
We should be testing the status code instead, not a response body property:
```
pm.test("Status code is 400", function () {
    pm.response.to.have.status(400);
});
```

nissimsan:
Just submitted a PR which removes `code` from the schema. We should still keep this issue open until we have replaced it with a status code assertion (or decided not to).

brownoxford:
At present, all of our definitions use a code that matches the http response code, so I'd agree that there is currently no added benefit to including this property. Adding a new `discussion` label so that we can keep this conversation going.

@OR13 @BenjaminMoe thoughts?

BenjaminMoe:
Agreed, it's redundant.

OR13:
I have seen the internal code system be used along side http... especially if you want to ever move the same api over a different transport, like web sockets. 

It's also useful if you throw a 400, and then provide a code that provides more detail in the body.

That being said, if we don't need the code now, we can remove it for now.

brownoxford:
I propose we discuss on the next call with the goal of moving to "ready-for-pr"

brownoxford:
This should be closed, we did this with #382

brownoxford:
Found a few more references to error code that need fixing.

PR #382, opened 2022-09-07
Removing code and message from responseSchema40x schemas

nissimsan:
This removes some of the asserted details from the responseSchema40x schemas: 
- `code` which seems redundant with the actual status code (see https://github.com/w3c-ccg/traceability-interop/issues/381).
- the precise error message, which is also redundant as it just describes the generic error http error.

### DIFF
--- a/docs/openapi/schemas/Error.yml
+++ b/docs/openapi/schemas/Error.yml
@@ -1,11 +1,7 @@
 type: object
 required:
-  - code
   - message
 properties:
-  code:
-    type: integer
-    format: int32
   message:
     type: string
   details:
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -4630,22 +4630,22 @@
 		},
 		{
 			"key": "responseSchema400",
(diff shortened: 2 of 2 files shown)


brownoxford (commented):
These values in the postman suite are auto-generated via GitHub workflow using the `tests/update_conformance_schemas.sh` script. Any desired changes should be made to the OpenAPI schema directly.

I would hold off on removing any `code` properties from response definitions pending the outcome of the discussion in https://github.com/w3c-ccg/traceability-interop/issues/381

I do not recommend removing the `message` property. This should be customizable with more specific pre-defined errors. For example, in [did.yml](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/resources/did.yml), we override the Bad Request message with something more specific.

While we do have the `details` parameter in error responses, that is a completely optional item that is left up to implementors to fill out with a helpful message. We added the 'message' property so that we could have some way of enforcing a more generic (but still helpful) error response text.

OR13 (approved):

brownoxford (changes requested):
These changes must be made in the OpenAPI schema yml files, not directly in the schemas embedded into the postman collection. Also note that you have removed the message overrides in addition to the code - we need to keep those in order to enforce the message text.

nissimsan:
I will update this.

nissimsan:
Updated now in the OAS and keeping the message.

brownoxford (approved):
LGTM

mkhraisha (approved):

PR #396, opened 2022-09-08
chore: require type property for credentialStatus

### LINKED ISSUE #395: Require type property in `options.credentialStatus`
The [vc-api states the following](https://w3c-ccg.github.io/vc-api/#issue-credential) regarding the `options.credentialStatus` object in the request body for `/credentials/issue`:

Screen Shot 2022-09-08 at 1 09 12 PM

In order for the `credentialStatus` object to be an object of the given form, the `type` property MUST be required.

Modify the existing OpenAPI spec to require the `credentialStatus.type` property.

brownoxford:
This PR makes the `credentialStatus.type` property required (when `credentialStatus` is present) in the `options` object when posting to `/credentials/issue`.

This change is driven by the vc-api specification as outlined in #395 

Fixes #395

### DIFF
--- a/docs/openapi/schemas/IssueCredentialOptions.yml
+++ b/docs/openapi/schemas/IssueCredentialOptions.yml
@@ -14,6 +14,8 @@ properties:
   credentialStatus:
     type: object
     description: The method of credential status.
+    required:
+      - type
     properties:
       type:
         type: string
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -2365,6 +2365,74 @@
 									"response": []
 								},
 								{
+									"name": "credentials_issue:options.credentialStatus.type:missing",
+									"event": [
+										{
+											"listen": "test",
+											"script": {
+												"exec": [
+													"pm.test(\"status code is 400\", function () {",
+													" pm.response.to.have.status(400);",
+													"});",
+													"",
+													"pm.test(\"response validates against schema\", function() {",
+													" const schemaString = pm.collectionVariables.get(\"responseSchema400\");",
+													" pm.response.to.have.jsonSchema(JSON.parse(schemaString));",
+													"});",
+													"",
+													""
+												],
+												"type": "text/javascript"
+											}
+										},
+										{
+											"listen": "prerequest",
+											"script": {
+												"exec": [
+													"let rawBody = pm.variables.get(\"rawBody\");",
+													"",
+													"// options.credentialStatus must contain \"type\" when present",
+													"rawBody.options.credentialStatus = {};",
+													"",
+													"// Request body must be serialized before sending over the wire.",
+													"pm.variables.set(\"requestBody\", JSON.stringify(rawBody));"
+												],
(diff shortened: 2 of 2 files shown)


brownoxford:
> Noting that this will be a big breaking change for us at least

I'd love to hear more about this one. Note that this is NOT making `credentialStatus` a required element, only that when `credentialStatus` is present, it MUST contain a `type` property. Previously, you could submit with an empty object for `credentialStatus`.

Are you saying that your current implementation omits `type`, or am I misunderstanding?

nissimsan (approved):

brownoxford:
@nissimsan @BenjaminMoe @OR13 Can you help me understand how this is a breaking change for you? That was unexpected, and I want to make sure I am not missing something here.

OR13 (changes requested):
I've like to see an explicit call out to all the apis related to this, and make sure we are ready to take this change, and we have updated everything consistently.

nissimsan:
@brownoxford , we need to review the VC-API, compare and agree on a common shape between interop and VC-API. Will you action this, pls?

OR13:
Provide links to the following:

- VCAPI Credential Issue
- VCAPI Credential Status Update

- Trace Credential Issue
- Trace Credential Status Update

The schemas MUST all agree.

The types, enums, etc... should be consistent.

mkhraisha:
Below are all the links, reformatting for legibility:

> * VCAPI Credential Issue

https://github.com/w3c-ccg/vc-api/blob/d275b08a7f44714b3e96e14808c9eb98718c7a5d/components/IssueCredentialOptions.yml#L29


> * Trace Credential Issue

https://github.com/w3c-ccg/traceability-interop/blob/1f2adae9ecc366222dc136185c1df78f7cc70bf5/docs/openapi/schemas/IssueCredentialOptions.yml#L14

Both of these are identical with the caveat that we require `'RevocationList2020Status'`


> * VCAPI Credential Status Update

https://github.com/w3c-ccg/vc-api/blob/d275b08a7f44714b3e96e14808c9eb98718c7a5d/issuer.yml#L70

> * Trace Credential Status Update

https://github.com/w3c-ccg/traceability-interop/blob/1f2adae9ecc366222dc136185c1df78f7cc70bf5/docs/openapi/resources/credential-status.yml#L23




Trace & VCAPI  Credential Status update match with both of them utilizing an array with properties:
`type`
`status`

the VCAPI however does not require any of them ( but they don't require any fields in general in their yaml, we already go further and require properties in trace, for example we require the type property to be type `'RevocationList2020Status'`

mkhraisha (changes requested):
This updates issueCredentialOptions, but does not modify the update credential status endpoint

brownoxford:
Next steps are to modify the PR to update the other use case for credentialStatus.type and make it required. This is a small enough change to add it to the PR.

brownoxford:
> This updates issueCredentialOptions, but does not modify the update credential status endpoint

@mkhraisha I believe this was covered in #366?

mkhraisha (approved):

Issue #377, opened 2022-09-06
Conformance test suite should support multi-tenant and dedicated tenant setups


nissimsan:
On a multi-tenant platform, `{{API_BASE_URL}}/did.json` does not work because the DID document sits under a tenant path. The conformance test suite must also support this. 

Possible solutions could be: 
1. Deducting the tenant path from `ORGANIZATION_DID_WEB`
2. Adding a tenant path as dedicated environment variable (github secret)

OR13:
This is related to the conversation we had regarding "how to discover the issuer"... TLDR... use did:web... but that means you need to implement support for did:web, using either:

1. `.well-known`... which limits you to "single tenant use cases".
2. `path based routes`... which work for both.

We currently only support 2, I think 1 should be avoided personally... since there are lots of cases where you might want the same origin to host multiple issuer dids... ( company divisions, inventory, locations, etc...)

nissimsan:
@brownoxford, FYI - let's continue solutioning here.

nissimsan:
Noting that this affects not only the `did_json` test, but all the ones depending on `getIssuerId`: 
https://github.com/w3c-ccg/traceability-interop/blob/main/tests/conformance_suite.postman_collection.json#L4369

BenjaminMoe:
`https://platform.transmute.industries/credentials/issue` is a site-wide resource

`https://platform.transmute.industries/organizations/org_mt46IVhChCDYEV6o/did.json` is a unique resource for an identifier

`https://platform.transmute.industries/organizations/org_mt46IVhChCDYEV6o/presentations/available` is an end point for a specific tenant

nissimsan:
Next step on this is a PR on the respec describing this

nissimsan:
@brownoxford, please note https://github.com/w3c-ccg/traceability-interop/pull/410 which addresses part of this issue. Specifically it fixes the `GET did_JSON` test case (1st in the conformance collection). 

I hope this helps clarify the issue?

We need to do the same for `credential_issuer_id` (and later also presentation endpoint). Here though, I think we should rather base is on an authenticated `/resolve`-call. This we can do either in the top level pre-request script, or with a couple of normal Postman calls. I would prefer the latter, but expect you probably don't, so want to hear your thoughts first before progressing further.

nissimsan:
We've addressed a lot of this.

PR #399, opened 2022-09-13
DID Discovery

nissimsan:
This introduces an authenticated DID discovery call before calling `/ did.json`. This enables support for multi tenant platforms. 

Fixes https://github.com/w3c-ccg/traceability-interop/issues/377


nissimsan:
We also need to fix `getIssuerId`. But let's do it stepwise and start from here.

BenjaminMoe (approved):

brownoxford (commented):
I would like to hold off on this change pending further discussion.

nissimsan:
@brownoxford, sure happy to hold off on this - it's important that we all get aligned on this.

BenjaminMoe:
- Single tenant hosing
- Multi tenant hosting 

We want to have a diagram around this in the respec document that covers the cases that we're interested in.

brownoxford:
Discussed on call, consensus is to update respec first, then continue discussion on this PR.

PR #400, opened 2022-09-13
wip: working on credentials/verify conformance

brownoxford:
This PR is the beginning of the work to add conformance testing for the `/credentials/verify` endpoint.

Right now, the problem I'm trying to solve is that mutating the credential properties in the VC results in the signature being invalid, causing negative tests to fail for one of two reasons - either the input fails validation before the signature is checked, or the signature checks fail.

We need for there to be a single reason for test failure, and have to eliminate the bad signature by coming up with a way to sign invalid credentials. We definitely don't want verification to succeed if a 3rd party presents a malformed credential with a valid signature.

Currently, I am able to do that by using the `@digitalbazaar/vc` package internals, bypassing much of the input validation code. I'm running into trouble with `id` and `type` fields, for which the JSON-LD parser definitely does not like invalid values.

If anyone has any suggestions for how to proceed, I would love to hear them.

My current plan is to pre-generate VCs for the test suite using the `generate-testdata.js` script. The commented-out test cases in that file represent the items with which I am having trouble.


brownoxford:
This is related to https://github.com/w3c-ccg/vc-api-verifier-test-suite/issues/26

tests/generate-testdata.js
```diff
@@ -0,0 +1,250 @@
...
+
+// suitePromise generates promise that resolves to an Ed25519Signature2018
+// signature suite implementation to use for issuing verifiable credentials.
+const suitePromise = (async () => {
+  const { Ed25519VerificationKey2018 } = await import('@digitalbazaar/ed25519-verification-key-2018');
+  const { Ed25519Signature2018 } = await import('@digitalbazaar/ed25519-signature-2018');
+
+  const keyPair = await Ed25519VerificationKey2018.generate();
+  keyPair.id = `did:key:${keyPair.fingerprint()}#${keyPair.fingerprint()}`;
+  keyPair.controller = 'did:key${keyPair.fingerprint()';
```
dlongley, line 15:
```suggestion
  keyPair.controller = `did:key${keyPair.fingerprint()}`;
```

tests/generate-testdata.js
```diff
@@ -0,0 +1,250 @@
...
+
+  {
+    const description = 'issuer:missing';
+    const document = klona(baseCredential);
+    delete document.issuer;
+    sampleVCs[description] = await addProof({ document });
+  }
+
+  {
+    const invalidValues = {
```
aljones15, line 177: BTW this looks like a good case for a map:

```js
const invaludValues = new Map([
  ['issuer:array', ['did:example:123']],
 ['issuer:boolean', false]
])
```

then you get an iterable by default

Also you might want to change the VC context here to allow for the invalid values

```js
  "issuer": {"@id": "cred:issuer", "@type": "@id"},
```

```js
        "issuer": {"@id": "cred:issuer", "@type": "xsd:Boolean"}
```

brownoxford: @aljones15 

When I modify `issuer` in the context in this way, and try to set `issuer: {"id": false}` in the credential, I still get the same error I get without modifying the context:
```
JsonLdError [jsonld.SyntaxError]: Invalid JSON-LD syntax; "@id" value must a string.
```
Note that setting `issuer: false` does not cause any errors, nor does `issuer: {"id": "did:web:example.com"}`

brownoxford: This appears to be working:
```
        context['@context'].VerifiableCredential['@context'].issuer = {
          '@id': 'cred:issuer',
          '@type': '@id',
          '@context': {
            id: 'xsd:boolean',
          },
        };
```

brownoxford: Mutating any `type` property or the `@context` to anything invalid continues to prove difficult.

brownoxford:
I'm moving this out of draft and making it ready for review. The current implementation provides a script that will regenerate a number of malformed (but validly signed) verifiable credentials for use with various bad request tests of the `/credentials/verify` endpoint.

Most mutations are able to be tested, but a number of mutations that change the `@context` or any `type` properties to bad values are not currently possible, largely because the RDF canonicalization algorithms simply will not work, and no signature can be generated.

tests/generate-testdata.js
```diff
@@ -0,0 +1,335 @@
...
+      const doc = klona(credential);
+      delete doc['@context'];
+      return doc;
+    };
+    sampleVCs.set(description, await addProof({ credentialMutator }));
+  }
+
+  {
+    const invalidValues = new Map([
+      // ['vc:@context:boolean', false],
```
mkhraisha, line 95: These are commented out because of the difficulty of testing with changes to context right?

tests/valid-credential.json
```diff
@@ -0,0 +1,9 @@
+{
+  "@context": ["https://www.w3.org/2018/credentials/v1"],
+  "credentialSubject": {
+    "id": "did:example:123"
+  },
+  "issuanceDate": "2006-01-02T15:04:05Z",
+  "issuer": "did:web:example.com",
```
mkhraisha, line 7: nit: This won't resolve (obviously), do we want to maybe use an issuer that would resolve as a valid example?

mkhraisha (approved):
There are a number of commented-out lines, which will make it difficult going forward for anyone to delete/modify because they will be scared to remove those lines. I think a better path forward is to remove those comments. 
Unless you believe a forthcoming PR that is reasonably soon will tackle those test issues.

tests/generate-testdata.js
brownoxford: Correct, any commented out test cases were ones I was not able to generate valid signatures for. If we can figure out how to do that in the future, that would be great.

tests/valid-credential.json
brownoxford: hmm. The goal of this request is to fail _in the right way_. If we think that an implementation might fail because the issuer cannot be resolved, then I would say yes, we should change this to something that will resolve. Suggestions welcome.

brownoxford: Agree 100% - we should use did:key for the issuer.

BenjaminMoe (approved):

tests/valid-credential.json
brownoxford: Issue #404 has been created to track this change.

tests/generate-testdata.js
dlongley: > Mutating any type property or the @context to anything invalid continues to prove difficult.

I just have to say ... that's great! :P ... the above is a sign of a good implementation / reasonably safe APIs. We want it to be hard to create invalid stuff -- since the only use case is a negative test suite.

TallTed: @dlongley -- Your turn to have an un-fenced `@context` in the quoted content that starts https://github.com/w3c-ccg/traceability-interop/pull/400#discussion_r975702456. Please edit to add code fences.

PR #409, opened 2022-09-21
fix: sync did schemas for service type property

### LINKED ISSUE #408: Update DidResolutionResponse `type` schema
PR #380 made some adjustments to the schema definitions of various `type` properties to restrict their values to "array of string" where previously a bare "string" value was allowed.

Among the changes, was modification of [`TraceabilityAPIDIDWebDocument.yml`](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/TraceabilityAPIDIDWebDocument.yml).

We need to ensure that the schema for [`DidResolutionResponse.yml`](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/DidResolutionResponse.yml) matches, otherwise all `did.json` files passing schema validation on direct access will fail validation when accessed through did web resolution.

brownoxford:
This PR ensures that the schemas related to DID documents are in sync with respect to the `type` property of `service` items, i.e., they must be array of string.

Fixes #408 
References #380

### DIFF
--- a/docs/openapi/schemas/DidResolutionResponse.yml
+++ b/docs/openapi/schemas/DidResolutionResponse.yml
@@ -17,7 +17,9 @@ properties:
             - serviceEndpoint
           properties:
             type:
-              type: string
+              type: array
+              items:
+                type: string
             serviceEndpoint:
               type: string
   didResolutionMetadata:
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -9026,7 +9026,7 @@
 		},
 		{
 			"key": "responseSchema200Identifiers",
-			"value": "{\"type\":\"object\",\"required\":[\"didDocument\"],\"properties\":{\"didDocument\":{\"type\":\"object\",\"required\":[\"service\"],\"properties\":{\"service\":{\"type\":\"array\",\"minItems\":1,\"items\":{\"type\":\"object\",\"required\":[\"type\",\"serviceEndpoint\"],\"properties\":{\"type\":{\"type\":\"string\"},\"serviceEndpoint\":{\"type\":\"string\"}}}}}},\"didResolutionMetadata\":{\"type\":\"object\"},\"didDocumentMetadata\":{\"type\":\"object\"}}}",
+			"value": "{\"type\":\"object\",\"required\":[\"didDocument\"],\"properties\":{\"didDocument\":{\"type\":\"object\",\"required\":[\"service\"],\"properties\":{\"service\":{\"type\":\"array\",\"minItems\":1,\"items\":{\"type\":\"object\",\"required\":[\"type\",\"serviceEndpoint\"],\"properties\":{\"type\":{\"type\":\"array\",\"items\":{\"type\":\"string\"}},\"serviceEndpoint\":{\"type\":\"string\"}}}}}},\"didResolutionMetadata\":{\"type\":\"object\"},\"didDocumentMetadata\":{\"type\":\"object\"}}}",
 			"type": "string"
 		},
 		{


mprorock (approved):
big thumbs up - assuming we want to properly support multiples

BenjaminMoe:
Are we making this assertion?

```
  "service": [
    {
      "id": "did:web:platform.transmute.industries:organizations:org_x5MMEJNOKvsrqnA5#traceability-api",
      "type": ["TraceabilityAPI"], <-- This should be an array
      "serviceEndpoint": "https://platform.transmute.industries/organizations/org_x5MMEJNOKvsrqnA5"
    }
  ]
}
```

I think we could say, "we expect the TraceabilityAPI service `type` to be declared in an array". Not sure if that is going to mesh with other specs that declare services.

https://identity.foundation/decentralized-web-node/spec/#service-endpoints
```
{
  "id": "did:example:123",
  "service": [{
    "id":"#dwn",
    "type": "DecentralizedWebNode",
    "serviceEndpoint": {
      "nodes": ["https://dwn.example.com", "https://example.org/dwn"]
    }
  }]
}
```

https://identity.foundation/didcomm-messaging/spec/#service-endpoint
```
{
    "id": "did:example:123456789abcdefghi#didcomm-1",
    "type": "DIDCommMessaging",
    "serviceEndpoint": [{
        "uri": "https://example.com/path",
        "accept": [
            "didcomm/v2",
            "didcomm/aip2;env=rfc587"
        ],
        "routingKeys": ["did:example:somemediator#somekey"]
    }]
}
```

Or are we making the assertion that we expect did's used in the conformance suite to have a specific structure?

brownoxford:
I believe that what we are saying is that we expect did:web documents accessed at `.../did.json` and through `/identifiers/did:web:...` to contain service entries that use `[]string` as the data type for service  type properties.

The intent of this PR is not to endorse that, but rather to bring the yaml in DidResolutionResponse in line with a decision already made in #380.

If we do not make this change, implementations will pass one of the following conformance tests, but not both:
* `identifiers`
* `did_json`

BenjaminMoe:
> This change is part of our ongoing process to prefer []string over oneOf(string,[]string)

I agree with the motivation. My main concern is that we can't make the assertion. For something like `trace-vocab`, we _can_ make the assertion that all of the documents defined within the specification have non-optional properties. In the case of dids, you can mix services in one did. And the following did would fail based on the schema check, right? 

```
  "service": [
    {
      "id": "did:web:platform.transmute.industries:organizations:org_x5MMEJNOKvsrqnA5#traceability-api",
      "type": ["TraceabilityAPI"], 
      "serviceEndpoint": "https://platform.transmute.industries/organizations/org_x5MMEJNOKvsrqnA5"
    },
   {
    "id":"#dwn",
    "type": "DecentralizedWebNode",
    "serviceEndpoint": {
      "nodes": ["https://dwn.example.com", "https://example.org/dwn"]
    }
  }
  ]
}
```

In this case my vote would go for `oneOf(string,[]string)`.

brownoxford:
> And the following did would fail based on the schema check, right?

Yes, this would fail conformance for both `did_json` and `identifiers` tests, both in the current versions as well as after the changes I'm proposing here.

Are you suggesting that we change both [`TraceabilityAPIDIDWebDocument.yml`](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/TraceabilityAPIDIDWebDocument.yml) and [`DidResolutionResponse.yml`](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/DidResolutionResponse.yml) to allow either `string` or `[]string` for `service` object `type` properties?

I'm not against that, and if we can get consensus would close this PR and open a new one after updating the details of the issue.

BenjaminMoe:
My proposal is:

- `string[]` in [TraceabilityAPIDIDWebDocument.yml](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/TraceabilityAPIDIDWebDocument.yml) - we can make the assertion we dont want optionality in our spec
- `oneof string,string[]` in [DidResolutionResponse.yml](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/DidResolutionResponse.yml) - we cant assert that will be true for all services that get resolved

And please double check me for consistency on this.

mkhraisha (approved):

brownoxford:
I've updated based on comments from @BenjaminMoe

BenjaminMoe (approved):

nissimsan (approved):

PR #410, opened 2022-09-22
did.json path determination from ORGANIZATION_DID_WEB

nissimsan:
This determines the location of the DID Document from `ORGANIZATION_DID_WEB`, rather than assuming it to be on `API_BASE_URL/did.json`. This supports multi-tenant situations (as well as potentially hosting the DID Doc elsewhere).

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -1,9 +1,9 @@
 {
 	"info": {
-		"_postman_id": "691a6d66-244e-4521-a0e5-6768dd6cf6f4",
+		"_postman_id": "802393fe-c7f8-48fc-93a3-85372c78d9c8",
 		"name": "Conformance Suite",
 		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
-		"_exporter_id": "4338127"
+		"_exporter_id": "15641111"
 	},
 	"item": [
 		{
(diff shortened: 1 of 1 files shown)


tests/conformance_suite.postman_collection.json
```diff
@@ -31,18 +31,43 @@
 										],
 										"type": "text/javascript"
 									}
+								},
+								{
+									"listen": "prerequest",
+									"script": {
+										"exec": [
+											"const convertDidToEndpoint = (did) => {",
```
brownoxford, line 39: If port is included in a did:web, the colon between host and port MUST be percent-encoded. This regex will need slight modification to account for that. The following is for reference, but I don't think it belongs here in a Postman request (will add more comments regarding this thought):

```javascript
const convertDidToEndpoint = (did) => {
  const re = /did:web:([a-zA-Z0-9.\-_%]+)(?::([a-zA-Z0-9.:\-_]+))?/;
  const match = did.match(re);
  if (!match) {
    throw new Error('DID is not a valid did:web');
  }
  const [host, path] = match.slice(1);
  const pathname = path?.split(':').join('/') || '.well-known';
  return `https://${decodeURIComponent(host)}/${pathname}/did.json`;
};
```

Note: would prefer not to encourage use of http here, even on localhost.

brownoxford (commented):
I would prefer to see this a part of the utility functionality in the pre-request script for the conformance folder.

Right now, we have a number of requests being fired off for `did.json` in order to get the issuer (via the `getIssuerId()` function), and we will probably have more in the future for getting the presentation endpoint.

It would be good to modify the Conformance folder pre-request script to add a private `convertDidToEndpoint(didweb)` and private `getDidJson(didweb)` function that makes use of `convertDidToEndpoint()` to fetch and cache did web documents so that we only need to request them once per run of the suite.

`getIssuerId()` could then be modified to use the fetch the cached did.json and extract the issuer from it instead of making a direct network request as it currently is doing.

A new public `getPresentationEndpoint()` function could be added to get the proper presentation endpoint in a similar way.

tests/conformance_suite.postman_collection.json
nissimsan: I stole this from here: 
https://github.com/transmute-industries/verifiable-data/blob/main/packages/did-web/src/convertDidToEndpoint.ts

nissimsan:
@brownoxford, thanks for the feedback. 

Only the `did_json` test case needs to get the did document without oauth security. So I intentionally made it this way, keeping it local to that test. 

`getIssuerId()` absolutely needs to support discovery too (I'm [painfully aware](https://w3c-ccg.github.io/traceability-interop/reports/conformance/)!) I actually had is working something like what you are describing, but intentionally reverted it;  I thought it would be nicer to do that with a "proper" authentication resolve call? This will minimize this somewhat hacky approach (splitting apart the did:web) to only where it is strictly needed.

tests/conformance_suite.postman_collection.json
brownoxford: This code will fail to convert a did:web DID with a port into a URL

brownoxford:
I would argue that the purpose of these conformance tests is to test each endpoint in isolation, and we should use the contents of `did.json` as data input, not a second test request.

If we have endpoints that require information from the did document, e.g., presentation endpoint or credential issuer id, we should fetch that data in a way that is least likely to fail (direct access of the `did.json` ) rather than using another endpoint to retrieve that data.

If any conformance test depends on ANOTHER endpoint being conformant (the `/identifiers/:did` endpoint), the tests will be less focused and could report false failures in the test matrix.

tests/conformance_suite.postman_collection.json
```diff
@@ -31,18 +31,43 @@
...
+											"  const regex = new RegExp(",
+											"    `did:web:(?<host>[a-zA-Z0-9/.\\\\-_]+)(:*)(?<port>[0-9]+)*(:*)(?<path>[a-zA-Z0-9/.:\\\\-_]*)`",
```
brownoxford, line 41:
```suggestion
											"    `did:web:(?<host>[a-zA-Z0-9/.\\\\-_]+)(?:%3A(?<port>[0-9]+))?(:*)(?<path>[a-zA-Z0-9/.:\\\\-_]*)`",
```

brownoxford (changes requested):

nissimsan:
@chris to make an alternative PR for this. Then close this.

nissimsan:
https://github.com/w3c-ccg/traceability-interop/issues/426

nissimsan:
Keeping this for a little longer for inspiration

brownoxford:
Closing in favor of #435

PR #414, opened 2022-09-22
chore: differentiate between 400 errors

### LINKED ISSUE #406: Differentiate credentials/verify 400 errors
Currently, the `/credentials/verify` endpoint returns a 400 error for bad requests. This should be made more specific so that there are two types of defined bad requests (both with 400 codes).

One type should indicate that the presented verifiable credential does not conform to the schema, and the second type should indicate that the provided proof signature is invalid.

brownoxford:
This PR introduces two new specific error responses for use when the request body is not conformant to the expected schema, and for when the request body contains an invalid signature.

Fixes #406

### DIFF
--- a/docs/openapi/resources/credential-verifier.yml
+++ b/docs/openapi/resources/credential-verifier.yml
@@ -29,7 +29,16 @@ post:
           schema:
             $ref: "../schemas/VerificationResult.yml"
     '400':
-      $ref: '../responses/BadRequest.yml'
+      description: |
+        The request was invalid, either because the request body did not conform
+        to the required schema, or because the request contained an invalid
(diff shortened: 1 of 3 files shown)


nissimsan (approved):

OR13 (approved):

brownoxford:
This should actually return a 2xx with a failed verification response, this is not a client error

nissimsan:
@brownoxford , pls add note from the call

OR13 (changes requested):
invalid signature is a 200 status response... 400 is reserved for malformed questions.

brownoxford:
Closing in favor of #431

Issue #351, opened 2022-08-16
Should options.created in `/credentials/issue` requests be tested as a date?


brownoxford:
Currently, [`IssueCredentialOptions.yml`](https://github.com/mesur-io/traceability-interop/blob/main/docs/openapi/schemas/IssueCredentialOptions.yml) defines `created` as a string representing "Date the proof was created".

Should we be adding negative testing that fails if this is not a valid date/time string (we do this already for `credential.issuance_date` for `/credentials/issue`)? If so, we should also modify the OpenAPI description text for this item to indicate that "This value MUST be an XML Date Time String" ([as we do for `credential.issuance_date`](https://github.com/mesur-io/traceability-interop/blob/main/docs/openapi/schemas/Credential.yml)).

nissimsan:
Discussion on the call, whether `created` should rather be provided by the server at issuance.

nissimsan:
Action: Double check with VC-API. If our decision, don't include this in the request options.

nissimsan:
https://w3c-ccg.github.io/vc-api/issuer.html#tag/Credentials

brownoxford:
The [current vc-api schema](https://github.com/w3c-ccg/vc-api/blob/main/components/IssueCredentialOptions.yml) specifies that `created` is an _optional_ element which defaults to the current system time. As such, we should remove it from our schema and disallow manual specification of the proof creation time.

brownoxford:
@OR13 pointed out that this would prevent VC (and VP) generation from being deterministic, and at the very least, would possibly cause issues for developers and/or testers expecting this functionality to be available.

After discussing on the call, consensus was that we should allow options.created in a request body for the `/credentials/issue` endpoint, and that we should implement testing for all relevant cases as outlined in #352.

brownoxford:
Original topic of this issue moved to #428

PR #425, opened 2022-09-27
feat: test proof created date within 10s

### LINKED ISSUE #352: Add positive testing for proof created date for `/credentials/issue` request
The `options.created` portion of a `/credentials/issue` request is an optional value indicating "Date the proof was created.". We should add positive conformance testing for the following:

- response `proof.created` should match request value if provided
- response `proof.created` should be close to "now" if request value is omitted

brownoxford:
This PR removes the negative testing related to bad values in `proof.created` (see #428), and adds positive testing to check that the `proof.created` value returned from a post to `/credentials/issue` is within 10 seconds of the current time.

Fixes #352
References #351

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -2219,84 +2219,6 @@
 									"response": []
 								},
 								{
-									"name": "credentials_issue:options.created:wrong_type",
-									"event": [
-										{
-											"listen": "test",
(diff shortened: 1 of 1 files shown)


mkhraisha (approved):
big +1 to these kinds of tests!

OR13 (approved):

nissimsan (approved):

PR #443, opened 2022-10-14
fix: update positive tests to use mutation

brownoxford:
This PR fixes a bug in the positive tests for `/credentials/issue` where a change to the way request body is mutated was not carried over from the negative tests.


nissimsan:
Great, @brownoxford !

Issue #468, opened 2022-11-15
Method for referencing stored verifiable credentials


brownoxford:
In order to be able to implement `/credentials/status`, a client needs to know what value to use in the `credentialId` POST body, and an implementation needs to be able to look up the associated credential in its internal storage.

Any `id` provided in the request body for a `/credentials/issue` request cannot be used by an implementation for security reasons - a bad actor could block use of ids by another client by flooding the system with requests using specially crafted `id` values. A careless actor could fail to use unique IDs

@brownoxford and @OR13 discussed offline and arrived at the following proposal:

**proposal**:
- When a client requests a revocable credential, if they provide an id or a credentialStatus they should receive a 400 error.
- When a revocable credential is issued the server should set the id and credentialStatus not the client.
- When a revocable credential has its status updated, the credential id should match the credential.id and the URL for the credential as a resource under the organization. (this last part might not be true for you if you use URNs)

Confirm all of this with a postman unit test.

*Note*: Implementations are responsible for generating a unique credential `id` when issuing a revocable credential, and clients are responsible for storing said `id` if they want to be able to later modify the credential through a request to `/credentials/status`.

brownoxford:
**Alternative Proposal (`storedCredential` property)**: This proposal is put forth to cover both the revocable credential scenario outlined above as well as other future scenarios where it may make sense to store a credential for later reference.

Add a new top-level property to the responses from `/credentials/issue` that contains the ID used to store the credential. This property would only be present if a credential is stored by the issuing implementation, and would most likely be _required_ if the credential being issued is revocable:

```json
{
  "verifiableCredential": ...
  "storedCredential": {
    "id": "urn:uuid:07aa969e-b40d-4c1b-ab46-ded252003ded"
  }
}
```

Under this proposal, it would NOT be an error to include `credential.id` in the request to `/credentials/issue` for a revocable credential. It would, however, still be an error to include `credential.credentialStatus` - this would be a good candidate for a separate issue as it does not relate to methods for referencing stored credentials.

ping @OR13

mprorock:
> Add a new top-level property to the responses from `/credentials/issue` that contains the ID used to store the credential. This property would only be present if a credential is stored by the issuing implementation, and would most likely be _required_ if the credential being issued is revocable:

This second approach seems like it could be handled separately and has a great degree of utility in many scenarios beyond revocation.

Proposal 1 seems like the best overall way of handling the revocation issue.

brownoxford:
[...]

PR #477, opened 2022-12-09
chore: make id required in credential/presentation

brownoxford:
This PR modifies the OpenAPI spec to make credential and presentation ID required. Associated examples, conformance testing, and helper scripts are also updated.

### DIFF
--- a/docs/openapi/schemas/Credential.yml
+++ b/docs/openapi/schemas/Credential.yml
@@ -1,6 +1,7 @@
 type: object
 required:
   - '@context'
+  - id
   - type
   - issuer
   - issuanceDate
--- a/docs/openapi/schemas/Presentation.yml
+++ b/docs/openapi/schemas/Presentation.yml
@@ -2,6 +2,7 @@ title: Presentation
 type: object
 required:
   - '@context'
+  - id
   - type
 properties:
   '@context':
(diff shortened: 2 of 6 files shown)


docs/openapi/schemas/Credential.yml
```diff
@@ -1,6 +1,7 @@
 type: object
 required:
   - '@context'
+  - id
```
nissimsan, line 4: @brownoxford , shouldn't this be on VerifiableCredential.yml (not Credential.yml)? 
Requiring this from the requestor breaks with both options 1 and 4 of https://github.com/w3c-ccg/traceability-interop/issues/468.

nissimsan (changes requested):

docs/openapi/schemas/Credential.yml
brownoxford: Regarding #468, this ticket is simply to bring our spec in line with the normative language that exists today - it is wholly separate from the changes that may come about from #468.

As for whether the ID should be required in Credential or VerifiableCredential yaml files - Based on responses from @OR13 and @mkhraisha, I have put the requirement on the base Credential.ym since VerifiableCredential is described as being a credential with an added proof and that normative statements should apply to both.

I'm expecting that we will need to make adjustments once #468 is resolved.

mkhraisha (approved):

docs/openapi/schemas/Credential.yml
nissimsan: Makes sense, @brownoxford

nissimsan (approved):

nissimsan:
@brownoxford , please resolve and merge offline

PR #480, opened 2022-12-12
chore: make verifications required in verification result

### LINKED ISSUE #469: Require `verifications` array in verification response
Action item from 2022-11-15 weekly meeting.

@OR13 suggested that simply including `verified` property in verification result is not very helpful in the case that verification failed.

> ... at the library implementation layer, often you will see an implementation provide details around why a verification failed. It could be because date was out of range, or because revocation was not resolvable, or credential was revoked. This is information that we could be providing, as a nice to have enhancement

@nissimsan pointed out that we do have some of this already in the form of the optional `verifications` property.

@brownoxford confirmed that the [current verification response](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/VerificationResult.yml) from verification of credentials and presentations only requires the `verified` boolean property to be present.

*Proposal:* make `verifications` array required in verification response.

### LINKED ISSUE #470: Update schema examples for VerificationResult
Schema for [VerificationResult](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/VerificationResult.yml) includes an entry where `title` has the value `Signature`. This is not one of the valid enum values from the [Verification](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/Verification.yml) schema.

brownoxford:
This PR makes the `verifications` array in the [VerificationResult.yaml](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/VerificationResult.yml) object required. The array can be empty, but it must be present.

Additionally, `Signature` has been added as an enum value for the `title` property of a `verifications` array item, `status` values `good` and `bad` have been changed to `Passed` and `Failed` respectively, and an entry for each of the possible `title` enum values has been added to the example verification result.

Fixes #469
Fixes #470

### DIFF
--- a/docs/openapi/schemas/Verification.yml
+++ b/docs/openapi/schemas/Verification.yml
@@ -4,15 +4,16 @@ properties:
   title:
     type: string
     enum: 
-      - Proof
       - Activation
       - Expired
+      - Proof
       - Revocation
+      - Signature
   status:
     type: string
     enum: 
-      - good
-      - bad
+      - Passed
+      - Failed
   description:
     type: string
 required: 
@@ -21,6 +22,6 @@ required:
 example:
   {
       "title": "Activation",
-      "status": "good",
+      "status": "Passed",
       "description": "This credential activated 2 weeks ago"
   }
--- a/docs/openapi/schemas/Verification.yml
+++ b/docs/openapi/schemas/Verification.yml
@@ -12,8 +12,8 @@ properties:
   status:
     type: string
     enum: 
-      - Passed
-      - Failed
+      - good
+      - bad
   description:
     type: string
 required: 
@@ -22,6 +22,6 @@ required:
 example:
   {
       "title": "Activation",
-      "status": "Passed",
+      "status": "good",
       "description": "This credential activated 2 weeks ago"
   }
--- a/docs/openapi/schemas/VerificationResult.yml
+++ b/docs/openapi/schemas/VerificationResult.yml
@@ -15,24 +15,24 @@ example:
     "verified": true,
     "verifications": [
         {
-            "status": "Passed",
+            "status": "good",
             "title": "Activation",
             "description": "This credential activated 2 weeks ago"
         },
         {
-            "status": "Passed",
+            "status": "good",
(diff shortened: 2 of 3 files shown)


docs/openapi/schemas/Verification.yml
```diff
@@ -4,15 +4,16 @@ properties:
   title:
     type: string
     enum: 
-      - Proof
       - Activation
       - Expired
```
TallTed, line 8: `Activation` should be paired with `Expiration` and `Revocation` ... 
or `Expired` should be paired with `Activated` and `Revoked`.

`Proof` and `Signature` might also be better tweaked to 
`Proved` and `Signed`...

brownoxford: Thanks for the feedback, @TallTed. I've just created #481 to discuss this sort of thing, would you mind commenting there so we can keep this focused on making the property required?

TallTed: Done

docs/openapi/schemas/Verification.yml
```diff
@@ -4,15 +4,16 @@ properties:
...
+      - Proof
       - Revocation
+      - Signature
   status:
     type: string
     enum: 
-      - good
-      - bad
+      - Passed
+      - Failed
```
nissimsan, line 16: This hasn't been discussed? Did you pick this out of Paul's example on https://github.com/w3c-ccg/traceability-interop/issues/469? 
While arguably marginally more human intuitive than `good`/`bad` it's an annoying little "paper cut" change, and I question if it's worth the pain.

nissimsan (changes requested):

docs/openapi/schemas/Verification.yml
brownoxford: I'm not tied to the change, reverted.

mkhraisha (approved):
Lgtm can we open an issue for good/bad vs the new language? I liked your proposal more

nissimsan (approved):

Issue #512, opened 2023-02-09
Bad Conformance Test `credentials_issue:credential.@context:invalid`


ipbyrne:
We have a conformance test that requires `https://w3id.org/traceability/v1` in its context. The `credentials_issue:credential.@context:invalid` test should be removed.

mkhraisha:
Why is this an invalid conformance test? 

I'm looking through the spec right now and I can't find the normative requirement to have `https://w3id.org/traceability/v1`  in the context but I believe it should be required to pass this test suite.

brownoxford:
This should not be an issue to remove the test, rather it should be an issue to modify the OpenAPI spec [Credential.yml](https://github.com/w3c-ccg/traceability-interop/blob/main/docs/openapi/schemas/Credential.yml) to remove the requirement for the URL to be in the context (modifying the test would be part of the process for resolving the issue).

```
properties:
  '@context':
    description: |
      The JSON-LD Context defining all terms in the Credential. This array
      MUST contain "https://w3id.org/traceability/v1".
    type: array
    items:
      type: string
```

That being said, I also don't fully understand the reason for removing this requirement.

brownoxford:
@ipbyrne The tests are currently implemented to test the requirements in the OpenAPI spec. Before we delete or modify tests, we need to have some discussion about whether or not the spec should be modified - if we agree it should be modified, then it should be modified and the relevant tests updated to reflect the new state of the spec.

mkhraisha:
I'm not sure whats going on with my earlier comment on this thread, but I'll repost the gist of it again: 

Why are we removing traceability vocab as a requirement from the traceability interop context?

brownoxford:
@ipbyrne The traceability interoperability specification extends the core Verifiable Credential API definitions described in the VC-API specification, so there will be normative requirements that are not present in the VC API specification (but which are still conformant to the VC API specification).

> you are making them choose between conforming to the VC API spec, or the custom trace version of it.

I disagree, a VC with `https://w3id.org/traceability/v1` in the `@context` is still conformant to the VC API specification.

brownoxford:
> If someone builds their API to conform to this requirement (enforcing the trace context being required), their API can only issue trace interop credentials, which means they can't issue any other ones.

This does seem like something we should evaluate. An implementation should be able to conform to the trace-interop spec without being prevented from issuing other types of credentials. @mkhraisha does this answer your concerns as well?

mkhraisha:
[...]

PR #519, opened 2023-02-13
Remove `credentials_issue:credential.id:missing` and `credentials_verify:vc:id:missing` test

ipbyrne:
This test requires an id to be present in the VC when it's sent to /credentials/issue but the VC-API spec implicitly states this field is not required (https://w3c-ccg.github.io/vc-api/#issue-credential), so we are again putting people in a position where their API can only conform to one or the other.

for: https://github.com/w3c-ccg/traceability-interop/issues/512#issuecomment-1428045434


OR13 (approved):
See:

- https://github.com/w3c/vc-data-model/issues/1031
- https://w3c-ccg.github.io/vc-api/#issue-credential

`credential.id` SHOULD be optional.

The response from the api is a `verifiable credential` which MUST have an `id`, according to the `trace-interop profile`, and MAY have an `id` according to the `vc-api` and the `vc-data-model`.

brownoxford:
@OR13 We decided that `credential.id` MUST be present in a request to `/credentials/issue`, see #468

OR13:
I realize that, but see the formal work items for the justification.

brownoxford:
> I realize that, but see the formal work items for the justification.

I don't fully follow you here, can you please elaborate?

OR13:
@brownoxford 

https://github.com/w3c-ccg/traceability-interop/pull/519#pullrequestreview-1295782357

https://github.com/w3c/vc-data-model/issues/1031

brownoxford (changes requested):
The request bodies for credential verification tests with bad requests are generated by the `update_conformance_vcs.js` script so that the mutated credentials can have a valid signature.

If we remove the `credentials_verify:vc:id:missing` test, we will also need to modify `update_conformance_vcs.js` so that it does not try to generate the signed request body for that test.

You will want to remove the following code block from that script:
```
  {
    const description = 'vc:id:missing';
    const credentialMutator = (credential) => {
      const doc = klona(credential);
      delete doc.id;
      return doc;
    };
    sampleVCs.set(description, await addProof({ credentialMutator }));
  }
```

brownoxford (approved):

mkhraisha (approved):

PR #522, opened 2023-03-01
fix: add request bodies to negative tests

brownoxford:
This PR ensures that negative authentication and authorization tests only fail for one reason. There were several tests that were not updated after we implemented request body mutation, and these were still trying to reference old (removed) variables resulting in an empty request body.

Other tests (presentation related) were not sending any request body at all and have been upgrade to start using request body mutation with the current behavior to send an empty JSON object.

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -6534,7 +6534,10 @@
 											"script": {
 												"exec": [
 													"// Request body must be serialized before sending over the wire.",
-													"pm.variables.set(\"requestBody\", JSON.stringify(pm.variables.get(\"rawBody\")));"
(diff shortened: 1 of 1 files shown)


mprorock (approved):

brownoxford:
> Will these changes make any current tests start failing? I am a little confused as to what these changes are doing. Is this just for cases where an API would return a 400 error before a 403 so you can't test the 403 response?

That's exactly it. Our implementation currently does some JSON validation on input before checking auth, and we return 400 here instead of 401/403. Understand that our implementation may need updating, but this exposed the fact that these tests which I am modifying send requests that are broken in more than one way (bad auth and bad request body). There should only be one conformance violation per test, so these changes fix the negative auth test requests to include a valid JSON request body.

PR #527, opened 2023-03-07
test: remove legacy context tests

### LINKED ISSUE #526: Stop requiring trace context in responses
The existing positive tests for `/credentials/issue` contain legacy tests from when we required `https://w3id.org/traceability/v1` to be present in the `@context` of a response body:

```
// Current OpenAPI v3.x schema cannot require specific array contents
pm.test("response @context contains traceability/v1", function() {
 const context = pm.response.json().verifiableCredential["@context"];
 pm.expect(context).to.contain("https://w3id.org/traceability/v1");
});
```

As @OR13 pointed out in #523, this test will fail for GS1 credentials, which have their own `@context` for traceability terms. These legacy tests should be removed from the Postman conformance suite.

brownoxford:
This PR removes the legacy conformance tests that were requiring issued VCs to contain `https://w3id.org/traceability/v1` in `@context`.

Fixes #526


mkhraisha (approved):

BenjaminMoe (approved):

BenjaminMoe:
Not needed on presentations, but might be required on traceable workflows.

OR13 (approved):

PR #569, opened 2023-07-03
removes requirement of messages to be a specific enum

mkhraisha:
We had a prior agreement on a call, that we should specify the shape of the errors, but we do not need to specify the exact value of the error message.

### DIFF
--- a/docs/openapi/responses/BadRequest.yml
+++ b/docs/openapi/responses/BadRequest.yml
@@ -6,5 +6,4 @@ content:
         - $ref: '../schemas/Error.yml'
         - type: object
           properties:
-            message:
-              enum: ["Bad Request: Your request body does not conform to the required schema"]
+            message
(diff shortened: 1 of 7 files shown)


BenjaminMoe (approved):

BenjaminMoe:
1. Is this valid syntax. For `message` i would expect `type: string` with an `example`
2. Do we want to use `allOf`? Seems like we want to have a fixed structure even if we don't have a specific message for the enum. 
3. 
```
description: Bad Request
content:
  application/json:
    schema:
      allOf:
        - $ref: '../schemas/Error.yml'
        - type: object
          properties:
            message
```

brownoxford (changes requested):
@BenjaminMoe is on the right track here. All of these errors in `docs/openapi/responses` that make use of `allOf` are overriding the `message` property of `Error.yml` to change it from `string` to `enum`. The `allOf` is no longer needed.

mkhraisha:
@brownoxford @BenjaminMoe lmk if that fixes it

brownoxford (approved):
Thanks Mahmoud!

nissimsan (approved):

PR #573, opened 2023-07-09
feat: change vc-format header to content-type

rhofvendahl:
`Content-type` is a standard header, and fits this use case.

### DIFF
--- a/docs/openapi/parameters/header/contentType.yml
+++ b/docs/openapi/parameters/header/contentType.yml
@@ -1,4 +1,4 @@
-name: vc-format
+name: content-type
 in: header
 required: true
 description: A verifiable credential format
--- a/docs/openapi/parameters/header/accept.yml
+++ b/docs/openapi/parameters/header/accept.yml
@@ -1,8 +1,8 @@
-name: content-type
+name: accept
 in: header
 required: true
 description: A verifiable credential format
 example: "vc"
 schema:
   type: string
-  enum: ["vc", "vc-jwt"]
+  enum: ["application/vc+ld+jwt", "application/vc+ld+json"]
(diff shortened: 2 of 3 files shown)


brownoxford (approved):

brownoxford (changes requested):
Actually, do we need to update the `enum` for `content-type`?

We may also want to create separate issues for:
* Updating relevant unit tests to include this header in requests
* Add negative unit testing for incorrect/missing values
* Update respec document with normative language describing the use of this header

rhofvendahl:
@brownoxford What would you suggest we update it to? On reflection it might be good to include `vc-cwt`, is that what you mean?

OR13:
See this thread on the accept header

https://github.com/vcstuff/draft-terbu-sd-jwt-vc/pull/118#issuecomment-1629782037

OR13:
POST /credentials/issue

Body = claimset (with the server reserving the right to override things)

--accept vc+ld+jwt

docs/openapi/parameters/header/contentType.yml
```diff
@@ -1,4 +1,4 @@
-name: vc-format
+name: content-type
```
OR13, line 1: change to the accept header.... use `application/vc+ld+jwt` and `application/vc+ld+json` for the enum

OR13 (changes requested):
see comments

nissimsan:
@rhofvendahl, pls consider comments.

docs/openapi/parameters/header/contentType.yml
brownoxford, line 1:
```suggestion
name: accept
```

OR13 (approved):

Issue #457, opened 2022-11-08
OAuth2 Authentication - Make scope a parameter for postman tests  


tsnyder-gs1us:
We noticed in the credentials issue postman collections (https://github.com/w3c-ccg/traceability-interop/blob/main/docs/tutorials/credentials-issue/credentials-issue.postman_collection.json) the Get Access Token Test has a hard coded value for scope (scope= resolve:dids issue:credentials). 

While in the Authentication Tutorial postman test (https://github.com/w3c-ccg/traceability-interop/blob/main/docs/tutorials/authentication/authentication.postman_collection.json) does not include the scope parameter.

Some OAuth2 providers require a specific scope value to be passed to get a token. Would it be possible to update the Traceability postman collections to accept scope as a parameter via the newman command line and remove the hard coded scope value?

nissimsan:
Hi @tsnyder-gs1us , 

We have struggled with the scope request in the past. It's an area which isn't properly standardized, so it's not surprising that this resurfaces. And we might need to account for provider nuances. 

It isn't perfectly clear which of the collections is causing you the problem:

- "_Some OAuth2 providers require a specific scope value to be passed to get a token_" sounds like it is the Authentication collection which should be added a request scopes. 
- "_remove the hard coded scope value_" sounds like it is "Credentials Issue" should have those scopes removed. 

I'm going to assume it is the latter, which seems most likely and closest relates to issues we've seen before. 

I fail to see how parameterization would help, though. If we moved those two scopes to be injected my newman, you would still end up failing the test. It would also be an anti-pattern to have the collection deviate from what is actually being tested.

However, I think it would be perfectly acceptable to remove `scope: "resolve:dids issue:credentials"` from the Credentials Issue collection _without_ adding spooky newman workarounds. This collection is about issuance, not API security - I actually think it's a mistake that we even do this. (API security is covered in great lengths in the Conformance suite). If you can confirm this would solve your problem, this should be a simple PR. 

@brownoxford, wouldn't you agree, we should just remove scope requests from interop tests? 

PS: You might also want to look at your oauth2 provider's configuration options - there might be a way you could tweak it to work.

tsnyder-gs1us:
We have confirmed that our OAuth2 Provider (Azure AD) requires a specific scope value to be passed when getting a OAuth token. This is why we suggested making it an environment variable for the postman collection that would be passed to the collection via the newman command line when being executed.

[...]

PR #579, opened 2023-07-18
chore: remove scope from interop suite

### LINKED ISSUE #550: Interop tests need to be updated to account for scop changes
The interoperability test suite needs to be updated to take into account the changes made for #457 with PR #545

brownoxford:
This PR updates the `scope` body parameter to use an environment variable instead of hard-coded scope names per #457.

As a reminder, it is now assumed that OAuth providers will return *all* scopes granted to the client ID in use when no scope is named in the token request (the default behavior for many providers, e.g. Auth0), and that implementations which require a value to be present in the token request (e.g., Azure) must provide a value for the `CLIENT_SCOPE` environment variable.

Fixes #550

### DIFF
--- a/docs/tutorials/presentations-exchange-oauth/presentations-exchange-oauth.json
+++ b/docs/tutorials/presentations-exchange-oauth/presentations-exchange-oauth.json
@@ -86,7 +86,7 @@
 						},
 						{
 							"key": "scope",
-							"value": "resolve:dids submit:presentations",
+							"value": "{{VERIFIER_CLIENT_SCOPE}}",
 							"type": "text"
 						}
 					]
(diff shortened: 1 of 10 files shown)


BenjaminMoe (approved):

docs/tutorials/presentations-exchange-oauth/presentations-exchange-oauth.json
```diff
@@ -86,7 +86,7 @@
 						},
 						{
 							"key": "scope",
-							"value": "resolve:dids submit:presentations",
+							"value": "{{VERIFIER_CLIENT_SCOPE}}",
```
OR13, line 89: we should review if this is still necessary.

PR #597, opened 2023-10-03
chore: add test for duplicate issuer value

### LINKED ISSUE #539: Duplicate JSON Keys  in request
by duplicate keys I mean: 
```
{
 "a": "test string 1",
 "a": "test string 2"
}
```

What should the issuer do when presented a credential to issue with duplicate keys?
JSON spec is silent on this: https://www.rfc-editor.org/rfc/rfc8259#section-4
JSON-LD spec says no duplicate keys: https://www.w3.org/TR/json-ld11/#json-ld-grammar

1. return a 400 bad request
2. use the second value, i.e. "test string 2"

Most JSON/JSON-LD libraries use option 2 when processing JSON input

brownoxford:
This PR adds a negative conformance test for duplicate JSON key `issuer` in a credentials issue request. The expectation is that the issuer will return a 400 error response.

This test is added into a new sub-folder called "Duplicate Keys" because any tests like this will differ from our standard tests in that it is not a simple mutation of the minimal request JSON object (you can't add duplicate keys in JSON). If we add more tests here,  each one will have a request body in raw text (not JSON) that includes the minimal credentials issue request with the appropriate key duplicated.

When we discuss this on call, we should determine whether or not we feel we need duplicate key tests for all keys.

Fixes #539

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -425,64 +425,6 @@
 							"name": "Bad Request",
 							"item": [
 								{
-									"name": "Duplicate Keys",
-									"item": [
-										{
-											"name": "credentials_issue.credential.issuer:dup",
-											"event": [
-												{
-													"listen": "test",
-													"script": {
-														"exec": [
-															"pm.test(\"status code is 400\", function () {",
-															" pm.response.to.have.status(400);",
-															"});",
-															"",
-															"pm.test(\"response validates against schema\", function() {",
-															" const schemaString = pm.collectionVariables.get(\"responseSchema400\");",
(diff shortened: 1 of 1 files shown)


nissimsan (approved):

BenjaminMoe (approved):

brownoxford:
This should be modified to EITHER return 400 OR make sure that the last provided value for the duplicate key is the one that is used in the returned credential.

mkhraisha (changes requested):
Would love it if this accommodate the standard JSON behavior to use the latest duplicate value rather than to error out.

mkhraisha (approved):

PR #603, opened 2023-10-17
remove JWS2020 reference

### LINKED ISSUE #562: Remove references to JsonWebSignature2020
We should remove all references to JsonWebSignature2020 as that work item is being withdrawn.

https://github.com/w3c/vc-jws-2020

mkhraisha:
fixes #562 

The agility tutorial itself does not actually use JWS2020, and currently only tests against a single signature suite.
do we want to maybe remove the entire tutorial instead?


brownoxford (changes requested):
@mkhraisha It looks like a generated artifact has been included here (the index.html), as well as `run.sh`, which contains a client secret. I would want to see those removed, and you may wish to rotate that secret.

The motion of removing this tutorial entirely should be made in a separate issue.

mkhraisha:
oops, will close this PR and rotate, reopening later

Issue #574, opened 2023-07-10
Update relevant unit tests to include "Accept" header in requests


rhofvendahl:
Currently the Accept header is used in Issue and Verify requests to specify the format of the verifiable credential. Legitimate values are vc (for json VCs with linked data proofs), vc-jwt, and (pending discussion) vc-cwt.

rhofvendahl:
Use of Content-Type is now uncertain - see discussion on https://github.com/w3c-ccg/traceability-interop/pull/573

nissimsan:
Ping @rhofvendahl

rhofvendahl:
@OR13 when we discussed this in August I believe you recommended making an issue (https://github.com/w3c/vc-jose-cose-test-suite/issues/8) in https://github.com/w3c/vc-jose-cose-test-suite, so that these changes could be referenced by multiple projects. It's looking like there hasn't been any activity in that project in the past month, though.

Do you know if that project is still moving forward? If it isn't actively being worked on, I'd be happy to move ahead with implementing this in the trace-interop repo.

OR13:
Yes, prefer work to be done in a place where the working group can benefit

rhofvendahl:
So you do expect vc-jose-cose-test-suite project to move forward? If that's the case I suppose I should just take on the issue there, as there doesn't seem to be many active developers at this time.

rhofvendahl:
Still WIP

rhofvendahl:
I've broken this into two PRs: https://github.com/w3c-ccg/traceability-interop/pull/605 for updating Accept headers, and https://github.com/w3c-ccg/traceability-interop/pull/606 for adding a test for the requirement of Accept.

PR #605, opened 2023-11-06
Update Accept header to application/vc+ld+json

rhofvendahl:
This PR updates references to `Accept` header in "Create Credentials" requests to use either `application/vc+ld+json` or `application/vc+ld+json+sd-jwt`, as discussed in issue https://github.com/w3c-ccg/traceability-interop/issues/574.

EDIT: using `application/sd-jwt` rather than `application/vc+ld+json+sd-jwt`, as discussed in 11/7/23 trace call.

### DIFF
--- a/docs/tutorials/vc-jwt/vc-jwt.collection.json
+++ b/docs/tutorials/vc-jwt/vc-jwt.collection.json
@@ -156,7 +156,7 @@
 				"header": [
 					{
 						"key": "Accept",
-						"value": "application/json",
+						"value": "application/vc+ld+json+sd-jwt",
 						"type": "text"
 					}
 				],
--- a/docs/tutorials/vc-jwt/vc-jwt.collection.json
+++ b/docs/tutorials/vc-jwt/vc-jwt.collection.json
@@ -156,7 +156,7 @@
 				"header": [
 					{
 						"key": "Accept",
-						"value": "application/vc+ld+json+sd-jwt",
+						"value": "application/sd-jwt",
 						"type": "text"
 					}
 				],
--- a/docs/openapi/parameters/header/accept.yml
+++ b/docs/openapi/parameters/header/accept.yml
@@ -5,4 +5,4 @@ description: A verifiable credential format
 example: "application/vc+ld+json"
 schema:
   type: string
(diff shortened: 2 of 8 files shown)


brownoxford (approved):

TallTed:
As we discussed on the call --

`application/vc+ld+json+sd-jwt` should not exist.

An `application/sd-jwt` is akin to an `application/zip`, and should not externally expose the media type of the document(s) it contains in its own media type; those should be revealed/discovered when that/those document(s) are extracted or sniffed.

In this scenario, an `application/vc+ld+json` document would be the result of extracting/sniffing from an `application/sd-jwt` document, and would be what is expected during such a transaction. Contents of other media types should lead to errors.

In sum: `Accept` headers in "Create Credentials" requests should be either `application/vc+ld+json` or `application/sd-jwt` (the latter of which will always package an `application/vc+ld+json`, in this context).

mkhraisha (approved):

rhofvendahl:
I'll be splitting out the data integrity proof updates, with sd-jwt work pending resolution in the jose-cose working group

docs/tutorials/vc-jwt/vc-jwt.collection.json
```diff
@@ -156,7 +156,7 @@
 				"header": [
 					{
 						"key": "Accept",
-						"value": "application/json",
```
OR13, line 159: I would remove this part

rhofvendahl: Could you say more? In this case the `Accept` header is on a `Verify` request, so just represents a preference that the verification response be in json format, right?

rhofvendahl:
@mkhraisha Once Orie's suggestion is resolved this should be ready to merge.

There are still some references to `application/sd-jwt` which may need to change if the jose-cose working group decides to go with something else, but there needs to be _some_ value in those cases so I think it should work for now.

OR13 (approved):

OR13:
Related PR: https://github.com/w3c/vc-jose-cose/pull/186

^ in case this PR is not merged, the current PR will continue to be correct.

PR #606, opened 2023-11-06
Add negative test for missing accept header

rhofvendahl:
This PR creates a test for credential creation which should fail if the server does not require an Accept header, as discussed in issue https://github.com/w3c-ccg/traceability-interop/issues/574.

NOTE: This will result in failed tests for all implementations which don't return `400` responses to requests without `Accept` headers.

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -6475,7 +6475,7 @@
 												"type": "text/javascript"
 											}
 										},
-                    {
+										{
 											"listen": "prerequest",
 											"script": {
 												"exec": [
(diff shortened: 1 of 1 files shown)


brownoxford (approved):

nissimsan:
If we agree to merge this, we need an issue to also add this to the respec.

mkhraisha (approved):

rhofvendahl:
It was decided on today's trace call that it does not make sense to return an error if a request Accept header is missing. An error should be returned if a media-type requested by the client is not available, but if the header is absent then the server should just return the default value.

PR #615, opened 2024-01-10
Purging didauth leftovers

nissimsan:

PR #618, opened 2024-01-16
remove did auth references in spec

### LINKED ISSUE #604: Remove DID Auth Endpoints from spec
We have previously resolved to not use DID Auth, we should remove references to DID Auth from the spec

mkhraisha:
in conjunction with #615 closes #604 

Still need someone to update section 10.

### DIFF
--- a/docs/spec/index.html
+++ b/docs/spec/index.html
@@ -243,16 +243,6 @@
     </p>
 
     <p>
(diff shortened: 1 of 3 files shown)


brownoxford:
We need to be careful today, if we merge this it will close #604.

brownoxford (approved):

Issue #482, opened 2023-01-03
Should we update the spec to use Status List 2021?


brownoxford:
Right now the [spec indicates that](https://w3c-ccg.github.io/traceability-interop/draft/#linked-data-signature-suites):

> Any system conforming with this specification for interoperability MUST support[ Revocation List 2020 ](https://w3c-ccg.github.io/vc-status-rl-2020/)to handle revocation of Verifiable credentials.

Is it time to change this to [Status List 2021](https://w3c-ccg.github.io/vc-status-list-2021/)?

This change was suggested by @mprorock, and I'd like to hear thoughts from @OR13 and @mkhraisha as draft editors.

mkhraisha:
I believe it is.

nissimsan:
Suggestion to let 2021 mature a bit further before taking this on.

OR13:
Assigned one of the editors of the work item: https://w3c.github.io/vc-status-list-2021/

nissimsan:
Let's do this together with VC2.0 upgrade.

rhofvendahl:
Let's do this together with VC2.0 upgrade.

PatStLouis:
Ignore my comment, I had the wrong link, correct link is [https://w3id.org/vc/status-list/2021/v1](https://w3id.org/vc/status-list/2021/v1) 

However the use of BitstringStatusList instead of the StatusList2021 sounds like it would be preferable with the VC2.0 upgrade?

mkhraisha:
Will tackle this soon, to change StatusList2021 to BitstringStatus list

mkhraisha:
This is complete

PR #628, opened 2024-03-08
Spec clean up

nissimsan:
- Removes empty Conformance section 
- Removes boilerplate requirements, not core to the mission of the spec

PR #629, opened 2024-03-12
change collection link to use the StatusList2021 collection

PatStLouis:
This PR is related to #482. The interoperability reporting GA currently tests the RevocationList2020, I would like to propose a change test revocation with StatusList2021 instead.

### DIFF
--- a/.github/workflows/interoperability-report.yml
+++ b/.github/workflows/interoperability-report.yml
@@ -348,7 +348,7 @@ jobs:
           token_endpoint: ${{secrets[format('{0}_TOKEN_ENDPOINT', matrix.actor)]}}
           api_base_url: ${{secrets[format('{0}_API_BASE_URL', matrix.actor)]}}
         run: |
-          npx newman run ./docs/tutorials/credentials-revocation/credentials-revocation.postman_collection.json \
+          npx newman run ./docs/tutorials/credentials-status-update/credentials-status-update.postman_collection.json \
           --env-var ORGANIZATION_DID_WEB=$organization_did_web \
           --env-var CLIENT_ID=$client_id \
           --env-var CLIENT_SECRET=$client_secret \


OR13 (changes requested):
I think its better to hold off on this, until we can support StatusList2021 and JWTs in these tests.

If you are willing to do that work, it would be greatly appreciated.

nissimsan (approved):

TallTed:
> [@OR13] in these tests

In *which* tests? Please link...

nissimsan:
The update corresponds with the path to the path to the collection. This is an oversight from whenever we updated the  collection. Even if outdated, this bug should still be fixed.

OR13:
@nissimsan fixing a broken path, to tests that do not match current requirements, does not help demonstrate conformance.

My change request, stands, please fix all of https://github.com/w3c-ccg/traceability-interop/tree/main/docs/tutorials/credentials-status-update 

Or don't register a test that is misleading.

PatStLouis:
There are conformance tests available for both `RevocationList2020` and `StatusList2021`. There are also interoperability tests available for both `RevocaitonList2020` and `StatusList2021`. However the [interoperability reporting matrix](https://w3c-ccg.github.io/traceability-interop/reports/interoperability/) uses the `RevocationList2020` tests for revocation interop. 

The current traceability interop spec states that:
> Any system conforming with this specification for interoperability MUST support [Status List 2021](https://w3c.github.io/vc-status-list-2021/) to handle revocation and status tracking for Verifiable Credentials

This PR is just to change which of the available test collections will be used for revocation interoperability reporting without additional work/tests needed. I would like to drop support for `RevocationList2020` in my implementation while still matching the interoperability requirements.

Regarding #628, there is mention of the data model 2.0, `BitstringStatusList` and JWT-VC. I would be happy to contribute to these test efforts in due time but is out of scope for my goal with this PR.

This being said, I'm also ok with leaving the tests as-is and keep support for `RevocationList2020` if need be.

OR13:
@PatStLouis Thanks for raising the PR. 

As you can see from https://github.com/w3c-ccg/traceability-interop/tree/main/docs/tutorials/credentials-status-update

There is more work to do, in order to support interop tests on this than just registering the test suite.

If you prefer to tackle the requirements support in a separate PR, I would also be fine modifying this one to remove the test that does not match requirements.

I also question the general utility of testing interop for this endpoint, since its an internal implementation detail.

I think the best solution in the short term would be to delete the test and associated content entirely.

cc @mprorock @mkhraisha

PatStLouis:
I believe having a revocation interoperability test is important as this is the best way to ensure each participant has implemented the `StatusList2021` properly (or whichever other status method is being implemented). If I'm only testing revocation against my own implementation, I might be able to verify the status of my own credentials without knowing I made a wrongful implementation, specially around the encoding/compression algorithm for the bitstring list. It's important to asses if `implementationA` can verify the status changes carried by `implementationB`, same way it's important to verify the proofs of issued VC across implementations. 

A prime example of this is the vc-api [issuer](https://w3c-ccg.github.io/vc-api-issuer-test-suite/) and [verifier](https://w3c-ccg.github.io/vc-api-verifier-test-suite/) interop test suite reports. Even if each implementation passes the tests, yet when I try to verify a credential I issue across a range of the verifiers, half of them return a failed verification. The main goal of interoperability should be to test implementation to implementation interactions.

There might be another way to achieve this without testing this endpoint directly, but until such a method exist I would vote to keep this in as this is the most practical way for the time being.

I will have a look at these 2 test collections to see the required changes to make the latter suitable for interop testing and assess if I have the bandwidth. 

Thank you for the comments/clarifications

PatStLouis:
@OR13 I had a comparison of both the revocation and status update collections and I can't see why the [revocation collection](https://github.com/w3c-ccg/traceability-interop/tree/main/docs/tutorials/credentials-revocation) is suitable for interop testing but the [status update collection](https://github.com/w3c-ccg/traceability-interop/tree/main/docs/tutorials/credentials-status-update) isn't. Could you provide more information as why you believe this to be the case?

OR13:
I don't think we should host tests for either here. The reason is that these are internal APIs, so we are simply forcing vendors to align on APIs interfaces which they do not actually need to be interoperable.

The get status endpoint does not interoperability, since verifiers occasionally so make network requests cross vendor to verify things.

OR13 (approved):

OR13:
I approved, but I hope we can remove everything that is not essential before the next round of testing.

OR13:
@PatStLouis I don't agree that we need to maintain tests for "internal endpoints"... because they are "internal".

I do agree that we need to test interoperability of credentials that have suspension and revocation abilities.

Vendor A presents suspended credential to Vendor B... what happens?
Vendor A presents revoked credential to Vendor B... what happens?

Vendor A might internally call "issue"
Vendor B might internally call "verify"

... but interop on these internal APIs is ... not relevant to the cross trust domain interop... and those internal APIs might be accomplished without ever implementing anything in HTTP... by testing internal APIs alongside external ones, we are muddying the waters on security, and creating more confusion.

Regardless, I approved this PR, because the place to fix is is a separate PR.

TallTed:
We have previously agreed that by design, there is no such thing as an "internal endpoint". Deployment might indeed put an endpoint inside a VPN or similar, such that authn and authz are unnecessary, but the *design* and *spec* have to include both authn and authz, such that the endpoint *can* be placed outside the VPN *and* supports both authn and authz *even internally*, where the deployer may indeed want to restrict use of that endpoint to specific users/clients which the deployer may define however they like.

mkhraisha (approved):

Issue #624, opened 2024-02-10
@context values can be objects according to the spec, conformance test-suite will raise this as a failed step


PatStLouis:
Two of the conformance negative tests will assert that context values can't contain `@vocab` objects. However in the [vc-data-model 1.1 spec](https://www.w3.org/TR/vc-data-model/#contexts) it states that:

> Subsequent items in the array MUST express context information and be composed of any combination of [URIs](https://www.w3.org/TR/vc-data-model/#dfn-uri) or objects.

Any reason for this misalignment from the vc-data-model 1.1 specification?

mkhraisha:
ping Patrick to cycle this issue, we have a PR open just pending feedback changes

mkhraisha:
This is no longer relevant, see #631

PR #630, opened 2024-03-12
Remove negative tests for context object items

PatStLouis:
These 2 tests are in contradiction with the VC 1.1 data model spec.

Addresses #624

### DIFF
--- a/tests/conformance_suite.postman_collection.json
+++ b/tests/conformance_suite.postman_collection.json
@@ -1163,75 +1163,6 @@
 									"response": []
 								},
 								{
-									"name": "credentials_issue:credential.@context.item.object",
-									"event": [
-										{
-											"listen": "test",
-											"script": {
-												"exec": [
-													"pm.test(\"status code is 400\", function () {",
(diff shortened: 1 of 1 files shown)


nissimsan (approved):
Seems like a good catch. Well done.

nissimsan:
@brownoxford, would love your input on this.

brownoxford:
In order to do this, we should first update the relevant schemas, as they currently specify array of string for `@context`.

brownoxford (changes requested):
In order to do this, we should first update the relevant schemas, as they currently specify array of string for `@context`.

TallTed:
Do these contradict or agree with VCDM 2.0?

PatStLouis:
@brownoxford I will look at including this change along in this PR

mkhraisha (approved):
assuming chris' objection is included

mkhraisha:
@PatStLouis any updates

PatStLouis:
closing since conformance testing will become out of scope for this specification

PR #654, opened 2024-04-25
Meeting cadence updated

nissimsan:


MEETINGS.md
```diff
@@ -3,7 +3,7 @@
 
 Meetings are held:
 
-- Tuesdays at [13:30 ET/10:30 PT](http://www.timebie.com/std/newyork.php?q=13.5)
+- Thursdays at [21:00CET/15:00 ET/12:00 PT](http://www.timebie.com/std/newyork.php?q=15)
```
TallTed, line 6: This should specify *one* timezone as the "anchor", with any others as "translations". There are several weeks each year where 21:00CET does *not* correspond to 15:00ET or 12:00PT. The timebie link here suggests ET is the "anchor", which suggests to me that ET should be the visible time, with a link to timebie such as below --
```suggestion
- Thursdays at 15:00 ET ([other timezones, accurate at time of viewing, which may vary with Daylight Saving changes](http://www.timebie.com/std/newyork.php?q=15))
```

It might be better to make the link go to the [CCG's W3C Calendar](https://www.w3.org/groups/cg/credentials/) (or the Traceability Task Force's, once it's been fixed).

mkhraisha (approved):

brentzundel (approved):

Issue #645, opened 2024-04-15
Add additional code generation languages


nissimsan:
https://w3c-ccg.github.io/traceability-interop/draft/#code-generation

Ideally we would have subsections with different code gen. languages, not only type script.

mkhraisha:
Personally not sure we should keep all of Section 9.1 in the spec at all.

mkhraisha:
Resolved to just remove section 9.1 from the spec.

PatStLouis:
Can you assign me please

nissimsan:
@PatStLouis confirmed assigned

mkhraisha:
@PatStLouis following up on this

PR #664, opened 2024-06-27
Remove section 9.1 from the spec

PatStLouis:
Addresses #645


brentzundel:
I think this was done in #661

nissimsan:
Dupe with #661

PR #674, opened 2024-10-10
feat: cgfr file

nissimsan:


nissimsan:
Administrative generation of CGFR static html

